PR #44, opened 2018-11-26
build(all): add renovate config

clebert:


renovate.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": ["config:base"],
+  "semanticPrefix": "build(all):",
+  "lockFileMaintenance": {"enabled": true},
+  "packageRules": [{"packagePatterns": ["^@types/"], "automerge": true}]
```
KingHenne: I think [this](https://renovatebot.com/docs/configuration-options/#automerge) applies here:

> Warning: GitHub currently has a bug where automerge won't work if a GitHub Organization has protected their master branch, and there is no way to configure around this. Hence, automerging will try and fail in such situations.

fahrradflucht: Do we risk upgrading to the typings for the next major of a dependency we didn't upgrade yet without noticing when doint this?

KingHenne: Had the same question, but the build task should catch this. On the other hand, I couldn't find whether renovate will wait for the CI to finish before merging. Also, see https://github.com/sinnerschrader/feature-hub/pull/44#discussion_r236404346

clebert: Ok i think we should deactivate automerge for now. i will make another commit.

KingHenne: No, you are right. This shouldn't necessarily lead to build errors, if for example we're not consuming the breaking part of the updated library.

fahrradflucht: It does (from my experience) - but there could be use cases where we don't use apis that broke but then introduce access later and ts doesn't catch it anymore because of the typings mismatch (though ideally unit-test should still catch)

KingHenne:
@clebert Do you want to mention that this closes #43?

clebert:
@KingHenne yeah good point. i mentioned it.

PR #77, opened 2018-11-28
docs(all): remove unnecessary double quotes

clebert:


KingHenne:
Do you consider the quotes unnecessary because the phrase is already linked?

PR #79, opened 2018-11-28
build(all): configure TS to use paths instead of project references

clebert:


package.json
```diff
@@ -6,14 +6,15 @@
     "packages/*"
   ],
   "scripts": {
-    "prebuild": "rimraf docs && lerna exec --parallel -- rimraf lib",
-    "build": "yarn build:sources && lerna exec --parallel -- typedoc --options ../../typedoc.js --out ../../docs/api/\\$LERNA_PACKAGE_NAME --readme README.md --tsconfig tsconfig.typedoc.json .",
+    "prebuild": "rimraf docs",
```
KingHenne: Deleting `lib` before building is still valid, even without project references, isn't it?

clebert: we delete it now before build:sources

clebert: before we didn't do that because we broke the caching. that doesn't exist anymore, does it?

KingHenne: ah, I missed that

PR #96, opened 2018-11-29
feat(examples): add TodoMVC Feature Apps

### LINKED ISSUE #82: Copy & Adapt: Feature Apps
Subtask of #29

KingHenne:
closes #82


examples/todomvc-main/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "@feature-hub/core": "^0.2.0",
+    "@feature-hub/react": "^0.2.0",
+    "@feature-hub/todomvc-todo-manager": "^0.0.0",
+    "styled-components": "^4.1.2"
+  },
+  "devDependencies": {
+    "@types/react": "^16.7.8",
+    "@types/styled-components": "^4.1.2",
+    "parcel-bundler": "^1.10.3",
+    "react": "^16.6.3",
```
clebert: why is react a dev dependency, but not lit-html in the header?

examples/todomvc-footer/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "@feature-hub/react": "^0.2.0",
+    "@feature-hub/todomvc-todo-manager": "^0.0.0"
+  },
+  "devDependencies": {
+    "@types/react": "^16.7.8",
+    "caniuse-lite": "^1.0.30000912",
+    "parcel-bundler": "^1.10.3",
+    "postcss-cssnext": "^3.1.0",
+    "postcss-modules": "^1.4.1",
+    "react": "^16.6.3",
```
clebert: why is react a dev dep?

examples/todomvc-main/package.json
KingHenne: It was originally because `react` was declared as external dependency in the webpack config. We should probably do the same with rollup.

examples/todomvc-footer/package.json
KingHenne: It was originally because `react` was declared as external dependency in the webpack config. We should probably do the same with rollup.

examples/todomvc-main/package.json
KingHenne: Well, [it's complicated](https://github.com/parcel-bundler/parcel/issues/144). I would suggest switching back to webpack.

KingHenne: We'll tackle this in #97.

Issue #97, opened 2018-11-30
Switch bundling of example feature apps from Parcel to Webpack to support externals


KingHenne:
see also https://github.com/parcel-bundler/parcel/issues/144

PR #104, opened 2018-11-30
build(examples): switch TodoMVC footer to webpack

KingHenne:
second part of #97

### DIFF
--- a/examples/todomvc-footer/webpack.config.js
+++ b/examples/todomvc-footer/webpack.config.js
@@ -1,6 +1,10 @@
-const path = require('path');
+// @ts-check
 
-module.exports = {
+const path = require('path');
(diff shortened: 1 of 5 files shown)


examples/todomvc-footer/webpack.config.js
```diff
@@ -0,0 +1,52 @@
+const path = require('path');
```
clebert: // @ts-check

examples/todomvc-footer/webpack.config.js
```diff
@@ -0,0 +1,52 @@
...
+
+module.exports = {
```
clebert: /** @type webpack.Configuration */
const config = {

PR #106, opened 2018-11-30
docs(server-renderer): fix introducing note

KingHenne:


fahrradflucht:
Wow - vscode rendered the previous markup as the correct link. That's pretty dangerous.

PR #107, opened 2018-11-30
test(examples): add unit tests for the todo manager

clebert:


examples/todomvc-todo-manager/src/__tests__/index.test.ts
```diff
@@ -0,0 +1,286 @@
...
+            },
+            allIds: ['0', '1', '2']
+          };
+
+          expect(todoManagerV1.getTodos()).toEqual(expectedTodos);
+        });
+      });
+
+      it('logs the added todo', () => {
+        todoManagerV1.add('test');
```
KingHenne: Just a note for a later refactoring: The add method should return the created Todo or its ID. Otherwise you can not create a Todo programmatically and set it to completed afterwards, which is something I wanted to do in the past.

packages/server-renderer/src/__tests__/index.test.ts
```diff
@@ -31,28 +31,32 @@ describe('#defineServerRenderer', () => {
...
-        featureServices: {},
-        config: {}
-      })['1.0'] as FeatureServiceBinder<ServerRendererV1>;
-
-      expect(
-        serverRendererBinder('test:1').featureService.serverRequest
-      ).toEqual(serverRequest);
+  describe('ServerRendererV1', () => {
+    describe('exposes', () => {
+      it('#serverRequest', () => {
```
KingHenne: change describe and it to a single `it('exposes a serverRequest')`

wallaby.js
```diff
@@ -4,9 +4,14 @@ module.exports = () => ({
   files: [
     'scripts/setup-test-framework.js',
     'packages/*/src/**/*.{ts,tsx}',
-    '!packages/*/src/**/*.test.{ts,tsx}'
+    '!packages/*/src/**/*.test.{ts,tsx}',
+    'examples/*/src/**/*.{ts,tsx}',
+    '!examples/*/src/**/*.test.{ts,tsx}'
```
KingHenne: simplify to `'{packages,examples}/*/src/**/*.{ts,tsx}'` etc.

PR #135, opened 2018-12-04
WIP: feat(server-renderer): move and adapt old SSR code

KingHenne:
This is the first iteration of the server renderer feature service.

Things to consider and/or fix:

- [ ] Can we avoid to use `setTimeout` in unit tests, and/or use `jest.useFakeTimers`?
- [ ] Move `rerenderWait` to config, remove default in `ServerRenderer` constructor.
- [ ] Add a test scenarios with multiple feature apps (with and without `rerender` calls).

Tasks for follow-up PRs:

- Handle a global render timeout.
- Add documentation.
- Integrate into `FeatureAppLoader`.
- Handle timeouts per loading feature app module.


packages/server-renderer/src/index.ts
```diff
@@ -12,21 +13,147 @@ export interface ServerRequest {
 
 export interface ServerRendererV1 {
   readonly serverRequest: ServerRequest | undefined;
+
+  startRendering(render: () => string): Promise<string>;
```
KingHenne: Maybe we can find a better name for this method.

packages/server-renderer/src/index.ts
```diff
@@ -12,21 +13,147 @@ export interface ServerRequest {
...
+  rerender(): Promise<void>;
 }
 
 export interface SharedServerRenderer extends SharedFeatureService {
   readonly '1.0': FeatureServiceBinder<ServerRendererV1>;
 }
 
+interface PromiseWithStatus<TResult> {
+  readonly promise: Promise<TResult>;
+  status: 'new' | 'pending' | 'final';
```
KingHenne: Is there maybe a more elegant approach to track whether we handled feature app module promises, and whether they are resolved/rejected?

packages/server-renderer/src/index.ts
```diff
@@ -12,21 +13,147 @@ export interface ServerRequest {
...
+}
+
+const undefinedRerenderErrorMessage =
+  'Invalid state: ServerRenderer#debouncedRerender is undefined.';
+
+export class ServerRenderer implements ServerRendererV1 {
+  private debouncedRerender?: () => Promise<void>;
+
+  private readonly registeredConsumers: (() => boolean)[] = [];
```
KingHenne: Is that a good name?

packages/server-renderer/src/define-server-renderer.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function defineServerRenderer(
+  serverRequest: ServerRequest | undefined
+): FeatureServiceProviderDefinition {
+  return {
+    id: 's2:server-renderer',
+
+    create: (env): SharedServerRenderer => {
+      const {rerenderWait = 50} = validateConfig(env.config) || {};
+      const serverRenderer = new ServerRenderer(serverRequest, rerenderWait);
```
clebert: rename to featureService?

packages/server-renderer/src/server-renderer-v1.ts
```diff
@@ -0,0 +1,78 @@
...
+  rerender(): Promise<void>;
+}
+
+export class ServerRenderer implements ServerRendererV1 {
+  private debouncedRerender?: () => Promise<void>;
+
+  private readonly consumerCompletedCallbacks: IsCompletedCallback[] = [];
+
+  public constructor(
+    public serverRequest: ServerRequest | undefined,
```
clebert: why is it not readonly?

packages/server-renderer/src/define-server-renderer.ts
KingHenne: I prefer the semantic name.

PR #159, opened 2018-12-07
fix(history-service): POP to initial location and replaced locations

### LINKED ISSUE #149: History Service: Back-navigation to initial location with browser-back is broken
Reason: https://github.com/ReactTraining/history/issues/502

KingHenne:
1.) We need to make sure that the initial browser location has a key.
2.) We need to update root history keys when the root location is replaced.

fixes #149

### DIFF
--- a/packages/history-service/src/base-history.ts
+++ b/packages/history-service/src/base-history.ts
@@ -5,6 +5,13 @@ export interface ConsumerHistory extends history.History {
   destroy(): void;
 }
 
+function setRootHistoryKey(
+  consumerLocation: history.Location,
+  rootHistoryKey: string | undefined
+): void {
+  consumerLocation.state = {...consumerLocation.state, rootHistoryKey};
+}
+
 export abstract class BaseHistory implements ConsumerHistory {
   protected readonly consumerHistory: history.MemoryHistory;
   protected readonly unregisterCallbacks: history.UnregisterCallback[] = [];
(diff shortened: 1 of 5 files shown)


packages/history-service/src/base-history.ts
```diff
@@ -148,4 +159,32 @@ export abstract class BaseHistory implements ConsumerHistory {
       rootHistoryKey: this.rootHistory.location.key
     };
   }
+
+  private updateRootHistoryKey(oldKey: string): void {
```
clebert: You can use `this.currentRootHistoryKey` instead of `oldKey` argument

packages/history-service/src/base-history.ts
```diff
@@ -148,4 +159,32 @@ export abstract class BaseHistory implements ConsumerHistory {
...
+  }
+
+  private handleRootLocationChange(
+    location: history.Location,
+    action: history.Action
+  ): void {
+    if (action === 'REPLACE') {
+      /* istanbul ignore else */
+      if (this.currentRootHistoryKey) {
+        this.updateRootHistoryKey(this.currentRootHistoryKey);
```
clebert: This method should be callable without arguments and should be named `updateAllRootHistoryKeys`

packages/history-service/src/base-history.ts
clebert: I had to think long and hard about all these auxiliary functions. I don't find them very understandable. For example: The same root location is accessed once via instance variable and once via location change handler. In addition, multiple locations are updated and not just one. My requested changes would have made my initial understanding easier.

KingHenne: I partially agree that this is not optimal yet. Why do you say that multiple locations are updated and not just one?

KingHenne: Also, I wanted to contain the handling of `currentRootHistoryKey` in the `handleRootLocationChange` method, i.e. checking for `undefined` and updating it. `updateRootHistoryKey` on the other hand is only responsible for calling `setRootHistoryKey` for the location that matches the `oldKey`.

PR #165, opened 2018-12-10
feat(core): separate configs for feature apps and services

### LINKED ISSUE #133: The `consumerConfigs` argument of the FeatureAppRegistry should be optional

clebert:
The configs for feature apps and services must now be provided separately by the integrator. This closes #133

### DIFF
--- a/packages/core/src/__tests__/feature-app-manager.test.ts
+++ b/packages/core/src/__tests__/feature-app-manager.test.ts
@@ -2,12 +2,11 @@ import {
   FeatureAppDefinition,
   FeatureAppManager,
   FeatureAppManagerLike,
-  FeatureAppModule,
-  FeatureServiceBindings,
-  FeatureServiceConsumerEnvironment,
   FeatureServiceRegistryLike,
+  FeatureServicesBinding,
   ModuleLoader
 } from '..';
+import {FeatureAppModule} from '../internal/is-feature-app-module';
 
 interface MockFeatureServiceRegistry extends FeatureServiceRegistryLike {
   registerProviders: jest.Mock;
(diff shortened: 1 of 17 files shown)


packages/core/src/__tests__/feature-service-registry.test.ts
```diff
@@ -250,20 +242,14 @@ describe('FeatureServiceRegistry', () => {
...
     });
 
     describe('for a feature service consumer with an id specifier and dependencies', () => {
       it('creates a bindings object with feature services', () => {
-        consumerConfigs = {
-          foo: {kind: 'consumerConfigFoo'}
-        };
-
-        registry = new FeatureServiceRegistry(consumerConfigs);
-
```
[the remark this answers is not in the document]
clebert: We don't need to test this method in conjunction with the ID specifier as it doesn't read the configuration.

packages/core/src/__tests__/feature-app-manager.test.ts
```diff
@@ -160,7 +160,12 @@ describe('FeatureAppManager', () => {
       );
 
       expect(mockFeatureAppCreate.mock.calls).toEqual([
-        [{config: mockConfig, featureServices: {}}]
+        [
+          {
+            config: mockConfig,
+            requiredFeatureServices: mockFeatureServicesBinding.featureServices
```
KingHenne: Sorry to be pedantic, but this would still lead to false positive tests, if fore some reason we would always set an empty object as `requiredFeatureServices` in the implementation. Additionally to this change, you also need to add a property to `mockFeatureServicesBinding.featureServices` in https://github.com/sinnerschrader/feature-hub/pull/165/commits/127c0b1c5045b9a23cea5d0d64ec20240d2e1f5e#diff-2aa666ad2c64b8a6beaf6877bd973747R32 (because of `toEqual`).

packages/history-service/src/index.ts
```diff
@@ -34,9 +34,13 @@ export interface SharedHistoryService extends SharedFeatureService {
   readonly '1.1': FeatureServiceBinder<HistoryServiceV1>;
 }
 
+export interface RequiredFeatureServices {
```
KingHenne: I like that we can type the required feature services now. I'm guessing you did the name change from `featureServices` to `requiredFeatureServices` because of the interface name (`RequiredFeatureServices`). I would prefer if we still called it `featureServices` though, despite the interface name. We don't really need the prefix there to discriminate them from some other feature services, don't you think?

packages/core/src/feature-app-manager.ts
```diff
@@ -1,31 +1,50 @@
...
-  extends FeatureServiceConsumerDefinition {
-  readonly ownFeatureServiceDefinitions?: FeatureServiceProviderDefinition[];
-
-  create(env: FeatureServiceConsumerEnvironment): TFeatureApp;
+export interface FeatureAppEnvironment<
+  TConfig,
+  TFeatureServices extends FeatureServices
+> {
+  /**
+   * A consumer config object that is provided by the integrator.
```
KingHenne:
```suggestion
   * A Feature App config object that is provided by the integrator.
```

packages/core/src/feature-service-registry.ts
```diff
@@ -14,19 +14,35 @@ export interface FeatureServices {
...
 
-export interface FeatureServiceBinding<TFeatureService> {
-  readonly featureService: TFeatureService;
-  unbind?(): void;
+export interface FeatureServiceEnvironment<
+  TConfig,
+  TFeatureServices extends FeatureServices
+> {
+  /**
+   * A consumer config object that is provided by the integrator.
```
KingHenne:
```suggestion
   * A Feature Service config object that is provided by the integrator.
```

PR #170, opened 2018-12-11
docs(all): providing config objects as an integrator

### LINKED ISSUE #164: Document Feature Hub configuration
We should improve the documentation around consumer configs, i.e. how exactly can the Integrator provide configs for certain consumers.

clebert:
Closes #164

### DIFF
--- a/README.md
+++ b/README.md
@@ -647,8 +647,9 @@ integrator, e.g.:
 
 #### Providing Config Objects
 
-The integrator can provide config objects for Feature Services and Feature Apps
-via the Feature Service registry and Feature App manager:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -639,6 +645,51 @@ integrator, e.g.:
 </section>
 ```
 
+#### Providing Config Objects
+
+The integrator can provide config objects for Feature Services and Feature Apps
+via the Feature Service registry and Feature App manager:
```
KingHenne: I just noticed: we should probably explicitly mention that the keys are the feature app/service IDs.

PR #174, opened 2018-12-13
feat(demos): rename integration-tests package to demos

clebert:


packages/demos/src/watch-demo.ts
```diff
@@ -0,0 +1,25 @@
+import {AddressInfo} from 'net';
+import {Configuration} from 'webpack';
+import {startServer} from './start-server';
+
+const demoName = process.argv[2];
+
+function loadWebpackConfig(): Configuration[] {
+  return (require(`./${demoName}/webpack-config`)
```
KingHenne: refactor `map`

PR #183, opened 2018-12-13
test(demos): add two initial itest scenarios for history service

clebert:


packages/demos/src/history-service/history-consumer.tsx
```diff
@@ -58,11 +65,15 @@ export const historyConsumerDefinition: FeatureAppDefinition<
   },
 
   create: env => {
-    const historyService = env.featureServices['s2:history'];
+    const {featureServices, idSpecifier} = env;
```
KingHenne: restructure in line above

PR #193, opened 2018-12-18
test(demos): add more history service itest scenarios

clebert:


packages/demos/src/history-service/index.test.ts
```diff
@@ -2,24 +2,64 @@
...
 
 jest.setTimeout(60000);
 
-async function getInputValue(selector: string): Promise<string | undefined> {
-  const element = await page.$(selector);
+class HistoryConsumerUI {
+  public static readonly a = new HistoryConsumerUI('a');
+  public static readonly b = new HistoryConsumerUI('b');
+
+  private constructor(private readonly prefix: 'a' | 'b') {}
```
KingHenne: Technically, based on your usage, this would be called a `suffix`. But we could also call it `specifier` for example.

KingHenne:
For some reason scenario 5 failed on Travis with Node 10.

PR #197, opened 2018-12-19
feat(history-service): remove history package type exports

clebert:
In addition, the package and file structure has been adapted to that of the other packages.


packages/demos/src/history-service/history-consumer.tsx
```diff
@@ -6,7 +6,7 @@ import {
   InputGroup,
   Label
 } from '@blueprintjs/core';
-import {History} from '@feature-hub/history-service';
+import {History} from 'history';
```
KingHenne: Should we add `@types/history` to the dependencies of `demos`?

clebert: done

PR #206, opened 2018-12-23
feat(core): make AsyncValue thenable

fahrradflucht:
When calling `getAsyncFeatureAppDefinition` in an async context it feels strange to have to await a special `promise` property and not beeing able to just await the function call.

Therefore make AsyncValue itself a thenable. This makes it possible to just
write `await manager.getAsyncFeatureAppDefinition('/example.js')`.

### DIFF
--- a/packages/core/src/async-value.ts
+++ b/packages/core/src/async-value.ts
@@ -1,9 +1,12 @@
-export class AsyncValue<TValue> {
+export class AsyncValue<TValue> implements PromiseLike<TValue> {
+  public then: Promise<TValue>['then'];
+
   public constructor(
     public readonly promise: Promise<TValue>,
     public value?: TValue,
     public error?: Error
   ) {
+    this.then = promise.then.bind(promise) as Promise<TValue>['then'];
     promise.then(val => (this.value = val)).catch(err => (this.error = err));
   }
 }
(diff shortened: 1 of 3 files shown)


packages/core/src/async-value.ts
```diff
@@ -1,9 +1,12 @@
-export class AsyncValue<TValue> {
+export class AsyncValue<TValue> implements PromiseLike<TValue> {
+  public then: Promise<TValue>['then'];
+
   public constructor(
     public readonly promise: Promise<TValue>,
     public value?: TValue,
     public error?: Error
   ) {
+    this.then = promise.then.bind(promise) as Promise<TValue>['then'];
```
fahrradflucht: This is necessary because TypeScript get tripped up with its inference chain by the `bind` call. I'm happy if someone has a better idea.

clebert:
I understand the motivation for this change. I don't like two things. We now need a special rule for TSLint, so it may happen that a "await" is forgotten. Furthermore we have two code paths. Strictly speaking the Promise should be tested additionally, not to test it speaks for knowledge of implementation details. What do you say to @KingHenne ? 

I guess we also need an extra test with async/await and try/catch. I'm not sure if this feature needs a catch method. Currently we use Jest reject method which is not a Real World Test.

KingHenne:
I had the same thoughts.

fahrradflucht:
Could we also just make the promise private? This would reduce the api surface and we wouldn't have to ways to do things, which we would then have to test...

fahrradflucht:
We decided against landing this PR because of tslint being unable to enforce no-dangling-promise on this thenable without explicitly setting it in the linter rules.

PR #212, opened 2019-01-02
fix(core): clarify error message

fahrradflucht:


packages/core/src/__tests__/feature-app-manager.test.ts
```diff
@@ -149,7 +149,7 @@ describe('FeatureAppManager', () => {
 
       expect(spyConsoleInfo.mock.calls).toEqual([
         [
```
clebert: Please add an extra test case, and maybe adapt the FeatureAppLoader also

PR #241, opened 2019-01-08
Meyer SSR

KingHenne:
As suggested by @stemey, it is sufficient for the server renderer to have a single `rerenderAfter` method (that takes a promise) instead of the `register` and `rerender` methods.


packages/server-renderer/src/config.ts
```diff
@@ -13,11 +13,9 @@ function isValidConfig(
     return false;
   }
 
-  const {rerenderWait} = config as ServerRendererConfig;
+  const {timeout} = config as ServerRendererConfig;
 
-  return (
-    typeof rerenderWait === 'undefined' || typeof rerenderWait === 'number'
-  );
+  return typeof timeout === 'undefined' || typeof timeout === 'number';
```
fahrradflucht:
```suggestion
  return ['undefined', 'number'].includes(typeof timeout);
```

KingHenne: thanks, but no thanks

packages/server-renderer/src/server-renderer-v1.ts
```diff
@@ -12,67 +18,51 @@ export interface ServerRendererV1 {
...
+      }
 
-      renderAndResolveIfCompleted();
-    });
-  }
-
-  private isRenderingCompleted(): boolean {
-    return this.consumerCompletedCallbacks.every(isCompleted => isCompleted());
+      return html;
+    } catch (error) {
```
fahrradflucht: this `try...catch` is unnecessary

packages/server-renderer/src/__tests__/index.test.ts
```diff
@@ -3,18 +3,23 @@ import {
   FeatureServiceBinder,
   FeatureServiceProviderDefinition
 } from '@feature-hub/core';
+// tslint:disable-next-line:no-implicit-dependencies
```
clebert: Its a test file, you can disable this rule for the whole file like in the other test files:

// tslint:disable:no-implicit-dependencies

packages/server-renderer/package.json
```diff
@@ -22,6 +22,9 @@
   "dependencies": {
     "@feature-hub/core": "^0.11.0"
   },
+  "devDependencies": {
+    "jest-mock-console": "^0.4.2"
```
clebert: maybe open an issue, so that we use this lib also in core and react

packages/server-renderer/src/define-server-renderer.ts
```diff
@@ -21,8 +21,10 @@ export function defineServerRenderer(
     id: 's2:server-renderer',
 
     create: (env): SharedServerRenderer => {
-      const {rerenderWait = 50} = validateConfig(env.config) || {};
-      const serverRenderer = new ServerRenderer(serverRequest, rerenderWait);
+      const {timeout} =
+        validateConfig(env.config) || ({} as ServerRendererConfig);
```
clebert: the or part should be handled by validateConfig

KingHenne: In my opinion the validate function should not change its input. Therefore, the default should be set here.

PR #243, opened 2019-01-09
feat(async-ssr-manager): rename from server-renderer

KingHenne:


packages/async-ssr-manager/src/define-async-ssr-manager.ts
```diff
@@ -0,0 +1,38 @@
...
+  ServerRequest
+} from './async-ssr-manager-v1';
+import {validateConfig} from './internal/validate-config';
+
+export interface SharedAsyncSsrManager extends SharedFeatureService {
+  readonly '1.0': FeatureServiceBinder<AsyncSsrManagerV1>;
+}
+
+export interface AsyncSsrManagerConfig {
+  timeout?: number;
```
clebert: why is that prop not readonly

packages/demos/package.json
```diff
@@ -5,12 +5,12 @@
   "license": "MIT",
   "dependencies": {
     "@blueprintjs/core": "^3.10.0",
+    "@feature-hub/async-ssr-manager": "^0.11.0",
```
clebert: 0.0.0

packages/history-service/package.json
```diff
@@ -20,8 +20,8 @@
   "module": "lib/esm/index.js",
   "typings": "lib/cjs/index.d.ts",
   "dependencies": {
+    "@feature-hub/async-ssr-manager": "^0.11.0",
```
clebert: 0.0.0

Issue #25, opened 2018-11-23
SSR: Implement `@feature-hub/async-ssr-manager` Feature Service


clebert:
To enable server side rendering, a Async SSR Manager Feature Service needs to be created. Feature Apps that want to be server side rendered must have this new Feature Service as a dependency.

**Tasks**

- [x] Add "old" SSR code into Server renderer package
- [x] Improve tests using Jest fake timers
- [x] Rename `server-renderer` package to `async-ssr-manager` #243 
- [x] Handle a global render timeout (config)
- [x] Integrate Async SSR Manager into `FeatureAppLoader` #248 
- [x] Write an integration test #251 
- [x] Write documentation #272 
- [ ] Add JSDoc comments #274
- [ ] Remove "incomplete" notice from README #274

PR #248, opened 2019-01-10
feat(react): add asyncSsrManager prop to FeatureAppLoader

KingHenne:
This is a task of #25.

### DIFF
--- a/packages/react/src/__tests__/feature-app-loader.test.tsx
+++ b/packages/react/src/__tests__/feature-app-loader.test.tsx
@@ -1,5 +1,6 @@
 // tslint:disable:no-implicit-dependencies
 
+import {AsyncSsrManagerV1} from '@feature-hub/async-ssr-manager';
 import {
   AsyncValue,
   FeatureAppDefinition,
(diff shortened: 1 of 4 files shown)


packages/react/src/__tests__/feature-app-loader.test.tsx
```diff
@@ -142,6 +154,18 @@ describe('FeatureAppLoader', () => {
         />
       );
     });
+
+    it('does not trigger a rerender on the Async SSR Manager', () => {
+      shallow(
+        <FeatureAppLoader
+          featureAppManager={mockFeatureAppManager}
+          src="/test.js"
```
clebert: maybe also add an serverSrc to ensure this does not trigger a rerender?

packages/react/src/__tests__/feature-app-loader.test.tsx
```diff
@@ -225,6 +249,20 @@ describe('FeatureAppLoader', () => {
       );
     });
 
+    describe('with an Async SSR Manager', () => {
+      it('does not trigger a rerender on the Async SSR Manager', () => {
+        shallow(
+          <FeatureAppLoader
+            featureAppManager={mockFeatureAppManager}
+            src="/test.js"
```
clebert: maybe also add an serverSrc to ensure this does not trigger a rerender?

packages/react/src/__tests__/feature-app-loader.test.tsx
```diff
@@ -225,6 +249,20 @@ describe('FeatureAppLoader', () => {
       );
     });
 
+    describe('with an Async SSR Manager', () => {
```
clebert: why does this describe statement does not exists in the test case above?

clebert: maybe remove it

PR #249, opened 2019-01-10
refactor(core): rename first argument of FeatureServiceRegistry constructor

clebert:


packages/core/src/__tests__/feature-app-manager.test.ts
```diff
@@ -15,7 +15,7 @@ interface MockFeatureServiceRegistry extends FeatureServiceRegistryLike {
 
 describe('FeatureAppManager', () => {
   let featureAppManager: FeatureAppManagerLike;
-  let mockRegistry: MockFeatureServiceRegistry;
+  let mockFeatureServiceRegistry: MockFeatureServiceRegistry;
```
KingHenne: I didn’t change the name of this variable because the property of the Festure App manager is still called `registry`.

clebert: Then we should rename the argument of the `FeatureAppManager` to `featureServiceRegistry`. Then it can stay that way. Is also only consistent.

KingHenne: Yes, maybe that is a better idea.

PR #268, opened 2019-01-14
feat(async-ssr-manager): remove the serverRequest property

### LINKED ISSUE #260: Server Request Provider: Extract request object from Async SSR Manager

KingHenne:
Instead the server request Feature Service should be used.

fixes #260

### DIFF
--- a/packages/async-ssr-manager/src/__tests__/index.test.ts
+++ b/packages/async-ssr-manager/src/__tests__/index.test.ts
@@ -1,23 +1,16 @@
 // tslint:disable:no-implicit-dependencies
 
-import {
-  FeatureAppEnvironment,
-  FeatureServiceBinder,
-  FeatureServiceProviderDefinition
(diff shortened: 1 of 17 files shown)


packages/async-ssr-manager/src/__tests__/index.test.ts
```diff
@@ -1,42 +1,30 @@
...
-
-    asyncSsrManagerDefinition = defineAsyncSsrManager(serverRequest);
   });
 
-  it('creates an Async SSR Manager definition', () => {
+  it('defines an id', () => {
     expect(asyncSsrManagerDefinition.id).toBe('s2:async-ssr-manager');
+  });
+
+  it('has no dependencies', () => {
```
fahrradflucht: shouldn't this be also done for optionals?

KingHenne: done

PR #270, opened 2019-01-15
docs(core): add jsdoc to FeatureServiceRegistry

### LINKED ISSUE #24: Core: Implement optional Feature Service dependencies
Currently Feature Apps cannot be loaded if not all dependencies can be registered and bound to the consumer. But not all dependencies are crucial for a Feature App to work.

**Remarks**

- Add a second optionalDependencies object as optional property to the `FeatureServiceConsumerDefinition`, to distinguish between required and optional dependencies.
- Handle optional dependencies in `registerFeatureServices` method of the `FeatureServiceRegistry`:
  - When topologically sorting, both sets must be joined
- Handle optional dependencies in `bindFeatureServices` method of the `FeatureServiceRegistry`:
  - The registry **must** not throw if an optional dependency is absent

**Tasks**

- [ ] Implementation
- [ ] Write documentation
- [ ] Add JSDocs
- [ ] Declare `s2:async-ssr-manager` as optional dependency in `s2:history`

fahrradflucht:
fixes #24

### DIFF
--- a/packages/core/src/feature-service-registry.ts
+++ b/packages/core/src/feature-service-registry.ts
@@ -128,17 +128,19 @@ export class FeatureServiceRegistry implements FeatureServiceRegistryLike {
    * Register a set of Feature Services to make them available for binding to
    * dependent consumers.
    *
-   * @throws Throws an error if dependencies of a registered service can't be
-   * fulfilled.
-   * @throws Throws an error if one of the registered services contains an
-   * invalid version according to semver notation.
+   * @throws Throws an error if the dependencies of one of the provider
+   * definitions can't be fulfilled.
+   * @throws Throws an error if one of the registered Feature Services contains
+   * an invalid version according to semver notation.
    *
-   * @param providerDefinitions Feature Services that should be registered.
-   * While the registry takes care of registering the provided definitions in
-   * the correct order, a Feature Service and its dependencies together or
-   * registering the dependencies earlier. It is not possible to provide
-   * dependencies later.
-   * @param consumerId The consumerId of the integrator itself.
+   * @param providerDefinitions Feature Services that should be registered. A
+   * Feature Service and its dependencies must either be registered together,
+   * or the dependencies must have already been registered. It is not possible
+   * to provide dependencies later. Sorting the provided definitions is not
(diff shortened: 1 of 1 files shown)


packages/core/src/feature-service-registry.ts
```diff
@@ -116,6 +124,22 @@ export class FeatureServiceRegistry implements FeatureServiceRegistryLike {
     private readonly options: FeatureServiceRegistryOptions = {}
   ) {}
 
+  /**
+   * Register a set of Feature Services to make them available for binding to
+   * dependent consumers.
+   *
+   * @throws Throws an error if dependencies of a registered service can't be
```
KingHenne: Maybe better: "Throws an error if the dependencies of one of the given provider definitions can't be fulfilled."

packages/core/src/feature-service-registry.ts
```diff
@@ -116,6 +124,22 @@ export class FeatureServiceRegistry implements FeatureServiceRegistryLike {
...
+   * fulfilled.
+   * @throws Throws an error if one of the registered services contains an
```
KingHenne: "Throws an error if one of the registered **Feature Services** contains..."

packages/core/src/feature-service-registry.ts
```diff
@@ -116,6 +124,22 @@ export class FeatureServiceRegistry implements FeatureServiceRegistryLike {
...
+   * invalid version according to semver notation.
+   *
+   * @param providerDefinitions Feature Services that should be registered.
+   * While the registry takes care of registering the provided definitions in
+   * the correct order, a Feature Service and its dependencies together or
+   * registering the dependencies earlier. It is not possible to provide
```
KingHenne: Please fix the second sentence.

packages/core/src/feature-service-registry.ts
```diff
@@ -116,6 +124,22 @@ export class FeatureServiceRegistry implements FeatureServiceRegistryLike {
...
+   * dependencies later.
+   * @param consumerId The consumerId of the integrator itself.
```
KingHenne: The ID of the consumer that provides the provider definitions.

PR #271, opened 2019-01-15
feat(async-ssr-manager): add rerender method

KingHenne:
Use case: When Feature Services want to trigger a rerender after their state has been changed, it would be a bit awkward to force them to pass `Promise.resolve()` into `rerenderAfter`.


KingHenne:
Closing because additionally providing the `rerender` method could suggest to consumers that this would be possible:

```js
const data = await fetch('https://example.com/foo');
asyncSsrManager.rerender();
```

Asynchronously triggering a rerender would not work, since all rerender promises must be registered synchronously during a render pass.

PR #272, opened 2019-01-15
docs(website): add SSR guide

KingHenne:


docs/guides/server-side-rendering-using-react.md
```diff
@@ -0,0 +1,161 @@
...
+
+> The `rerenderAfter` method must be called synchronously during a render pass,
+> since the Async SSR Manager synchronously checks after every render pass
+> whether there are rerender promises it needs to await, and then do another
+> render pass.
+
+### As Feature Service
+
+If a Feature Service consumer changes shared state of a Feature Service during a
+render pass on the server, the Feature Service can trigger a rerender to give
```
fahrradflucht: ~`can`~ `should`

packages/website/sidebars.json
```diff
@@ -9,6 +9,7 @@
       "guides/integrating-the-feature-hub",
       "guides/writing-a-feature-app",
       "guides/writing-a-feature-service",
+      "guides/server-side-rendering-using-react",
```
fahrradflucht: What do you think about moving this down at least one entry? I think sharing the browser history is more essential than ssr.

PR #287, opened 2019-01-21
feat(serialized-state-manager): create package

KingHenne:


packages/serialized-state-manager/README.md
```diff
@@ -0,0 +1,42 @@
+# @feature-hub/serialized-state-manager
+
+[![Package Version][package-badge]][package-npm]
+[![Website][website-badge]][website] [![API][api-badge]][api]
+
+A Feature Service to manage handing over state to the client when server-side
```
clebert: Alternative proposal for the description:
`A Feature Service for managing the state transfer to the client when rendering on the server.`

packages/serialized-state-manager/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@feature-hub/serialized-state-manager",
+  "version": "0.0.0",
+  "description": "A Feature Service to manage handing over state to the client when server-side rendering.",
```
clebert: See comment above.

clebert:
Also please add a link to the website sidebar.

packages/serialized-state-manager/README.md
KingHenne: I like it. 👍

packages/serialized-state-manager/package.json
KingHenne: > Also please add a link to the website sidebar.

I suggest we add the link when there’s more than empty methods that throw when called. 🙃

KingHenne:
> Also please add a link to the website sidebar.

I suggest we add the link when there’s more than empty methods that throw when called. 🙃

PR #291, opened 2019-01-23
feat(demos): add state serialization to async-ssr-manager demo

KingHenne:
In a follow-up PR Id' like to rename the demo from `async-ssr-manager` to `ssr`.

### DIFF
--- a/packages/demos/src/start-server.ts
+++ b/packages/demos/src/start-server.ts
@@ -4,16 +4,28 @@ import {Server} from 'http';
 import webpack from 'webpack';
 import devMiddleware from 'webpack-dev-middleware';
 
-export interface MainHtmlRendererOptions {
+export interface AppRendererOptions {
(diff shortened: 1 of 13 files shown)


packages/demos/src/start-server.ts
```diff
@@ -29,17 +42,19 @@ function createDocumentHtml(bodyHtml: string): string {
...
+  renderApp: AppRenderer | undefined,
   demoName?: string
 ): Promise<Server> {
   const port = await getPort(demoName ? {port: 3000} : undefined);
   const app = express();
 
   app.get('/', async (req, res) => {
     try {
-      const mainHtml = renderMainHtml ? await renderMainHtml({port, req}) : '';
+      const {html: appHtml = '', serializedStates = ''} = renderApp
```
clebert: please rework

PR #309, opened 2019-01-29
refactor(core): extract log and error message to functions

fahrradflucht:


packages/core/src/feature-service-registry.ts
```diff
@@ -120,6 +120,108 @@ function createUnsupportedFeatureServiceMessage(
...
+function alreadyRegisteredFeatureServiceMessage(
+  providerId: string,
+  consumerId: string
+): string {
+  return `The already registered Feature Service ${JSON.stringify(
+    providerId
+  )} could not be re-registered by consumer ${JSON.stringify(consumerId)}.`;
+}
+
+function successfullyBoundFeatureServiceMessage(
```
KingHenne: always start with `featureService`/`featureServices`

PR #310, opened 2019-01-29
build(all): include snapshot files in wallaby config

KingHenne:


KingHenne:
@fahrradflucht Could you please confirm that this fixes the problem for you as well?

PR #321, opened 2019-01-31
Externals Validator

fahrradflucht:


packages/core/src/externals-validator.ts
```diff
@@ -0,0 +1,53 @@
+import {coerce, satisfies} from 'semver';
+
+export interface Externals {
```
clebert: Rename to `ProvidedExternals` ?

PR #322, opened 2019-01-31
feat(demos): use addUrlForHydration for preloading in ssr demo

KingHenne:


packages/demos/src/server-side-rendering/integrator.tsx
```diff
@@ -19,6 +18,18 @@ function getSerializedStatesFromDom(): string | undefined {
   return (scriptElement && scriptElement.textContent) || undefined;
 }
 
+function getUrlsForHydrationFromDom(): string[] | undefined {
```
clebert: Please always return an empty array

KingHenne: and then check the length in the guard below? why?

KingHenne: the guard is only theoretical anyway, so maybe getting rid of it and returning an empty array here is indeed better

KingHenne: fixed (sorry, I accidentally amended)

PR #335, opened 2019-02-05
feat(async-ssr-manager): rename rerenderAfter to scheduleRerender

KingHenne:


packages/async-ssr-manager/src/__tests__/index.test.ts
```diff
@@ -163,7 +163,9 @@ describe('asyncSsrManagerDefinition', () => {
         it('rejects with an error after the configured timeout', async () => {
           const asyncSsrManager = asyncSsrManagerBinder('test').featureService;
           const mockRender = jest.fn(() => {
-            asyncSsrManager.rerenderAfter(new Promise<void>(() => undefined));
+            asyncSsrManager.scheduleRerender(
+              new Promise<void>(() => undefined)
```
clebert: <never>

packages/async-ssr-manager/src/internal/async-ssr-manager.ts
```diff
@@ -26,15 +26,15 @@ export class AsyncSsrManager implements AsyncSsrManagerV0 {
     return Promise.race([renderPromise, renderingTimeout(this.timeout)]);
   }
 
-  public rerenderAfter(promise: Promise<unknown>): void {
+  public scheduleRerender(promise: Promise<unknown>): void {
```
clebert: ad default parameter

PR #337, opened 2019-02-05
feat(async-ssr-manager): support re-scheduling rerender

KingHenne:


packages/async-ssr-manager/src/__tests__/index.test.ts
```diff
@@ -133,6 +133,82 @@ describe('asyncSsrManagerDefinition', () => {
...
+
+          let renderPass = 0;
+          let consumerResult = 0;
+
+          const mockRender = jest.fn(() => {
+            renderPass += 1;
+
+            if (renderPass === 1) {
+              (async () => {
+                const promise1 = new Promise<number>(
```
clebert: move to a named function

PR #338, opened 2019-02-05
docs(async-ssr-manager): update JSDocs comments and SSR guide

KingHenne:


packages/async-ssr-manager/src/index.ts
```diff
@@ -42,14 +39,36 @@ export interface AsyncSsrManagerV0 {
...
-   * synchronously checks after every render pass whether there are rerender
-   * promises it needs to await, and then do another render pass.
+   * Services. It schedules a rerender with an optional promise representing an
+   * asynchronous operation. The method must be called synchronously during a
+   * render pass, or while already scheduled asynchronous operations are
+   * running.
    *
-   * @param promise When this promise resolves, a rerender should be triggered.
+   * When no asynchronous operation is given, the method must be called
+   * synchronously during a render pass. So this will not work:
```
fahrradflucht: `This means that this **won't work**:`

packages/async-ssr-manager/src/index.ts
```diff
@@ -42,14 +39,36 @@ export interface AsyncSsrManagerV0 {
...
+   * asyncSsrManager.scheduleRerender(dataPromise);
+   * const data = await dataPromise;
+   * ```
+   *
+   * Calling it while already scheduled asynchronous operations are running,
+   * does not lead to multiple render passes, but instead the already scheduled
+   * rerender is deferred until every registered asynchronous operation has
+   * finished.
+   *
+   * @param asyncOperation The scheduled rerender shall be deferred at least
```
fahrradflucht: Rework this 😉

PR #341, opened 2019-02-06
feat(core): make ExternalsValidator optional for the FeatureAppManager

KingHenne:
When working on updating the documentation regarding externals, @fahrradflucht and I noticed, while reading the [Module Loader section](https://feature-hub.io/docs/guides/integrating-the-feature-hub#module-loader), that it probably makes more sense to make the `ExternalsValidator` optional, analogous to the optional module loader. Because only when an integrator uses the optional module loader to support loading Feature Apps via URL, it makes sense for him to also validate the externals that he provides for the separately deployed Feature Apps.

**Note:** Making the `ExternalsValidator` optional for the `FeatureServiceRegistry` as is done in a [follow-up PR](https://github.com/sinnerschrader/feature-hub/pull/343).

### DIFF
--- a/packages/core/src/__tests__/feature-app-manager.test.ts
+++ b/packages/core/src/__tests__/feature-app-manager.test.ts
@@ -175,6 +175,27 @@ describe('FeatureAppManager', () => {
       ]);
     });
 
+    describe('without an ExternalsValidator provided to the FeatureAppManager', () => {
+      describe('with a Feature App definition that is declaring external dependencies', () => {
+        beforeEach(() => {
+          mockFeatureAppDefinition = {
+            ...mockFeatureAppDefinition,
+            dependencies: {
+              externals: {
+                react: '^16.0.0'
+              }
+            }
+          };
+        });
+
+        it("doesn't throw an error", () => {
+          expect(() => {
+            featureAppManager.getFeatureAppScope(mockFeatureAppDefinition);
+          }).not.toThrowError();
+        });
+      });
+    });
+
     describe('with an ExternalsValidator provided to the FeatureAppManager', () => {
       beforeEach(() => {
         featureAppManager = new FeatureAppManager(mockFeatureServiceRegistry, {
@@ -204,10 +225,7 @@ describe('FeatureAppManager', () => {
 
         it('calls the provided ExternalsValidator with the defined externals', () => {
           try {
-            featureAppManager.getFeatureAppScope(
-              mockFeatureAppDefinition,
-              'testIdSpecifier'
-            );
+            featureAppManager.getFeatureAppScope(mockFeatureAppDefinition);
           } catch {}
 
           expect(mockExternalsValidator.validate).toHaveBeenCalledWith({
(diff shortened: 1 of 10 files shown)


packages/core/src/__tests__/feature-app-manager.test.ts
```diff
@@ -185,79 +175,107 @@ describe('FeatureAppManager', () => {
       ]);
     });
 
-    describe('with a Feature App definition that is failing the externals validation', () => {
-      let mockError: Error;
-
+    describe('with an ExternalsValidator provided to the FeatureAppManager', () => {
```
clebert: Don't we need another test that tests the declared externals to be ignored if no validator is configured?

packages/core/src/feature-app-manager.ts
```diff
@@ -70,8 +70,28 @@ export interface FeatureAppManagerLike {
...
+   * `@feature-hub/module-loader-commonjs` package).
+   */
   readonly moduleLoader?: ModuleLoader;
+
+  /**
+   * When using a {@link #moduleLoader}, it might make sense to validate
+   * external dependencies that are required by Feature Apps against the
+   * shared dependencies that are provided by the integrator. This makes it
+   * possible that an error is already thrown when creating a Feature App with
+   * incompatible external dependencies. This gives the author early feedback as
```
clebert: Is the author an entity that makes sense in the context of open source software?

clebert: Possibly replace author with integrator

packages/core/src/__tests__/feature-app-manager.test.ts
KingHenne: yes

packages/core/src/feature-app-manager.ts
KingHenne: how about "This enables early feedback as to whether a Feature App is compatible with the integration environment."?

packages/core/src/__tests__/feature-app-manager.test.ts
KingHenne: done

packages/core/src/feature-app-manager.ts
```diff
@@ -70,8 +70,28 @@ export interface FeatureAppManagerLike {
...
+   */
   readonly moduleLoader?: ModuleLoader;
+
+  /**
+   * When using a {@link #moduleLoader}, it might make sense to validate
+   * external dependencies that are required by Feature Apps against the
+   * shared dependencies that are provided by the integrator. This makes it
+   * possible that an error is already thrown when creating a Feature App with
+   * incompatible external dependencies. This enables early feedback as to
+   * whether a Feature App is compatible with the integration environment.
```
fahrradflucht: Maybe just omit `This makes it
   * possible that an error is already thrown when creating a Feature App with
   * incompatible external dependencies.`

KingHenne: How about "This makes it possible that an error is already thrown when creating a Feature App with incompatible external dependencies, and thus enables early feedback as to whether a Feature App is compatible with the integration environment."?

KingHenne: fixed

PR #343, opened 2019-02-07
feat(core): make ExternalsValidator optional for the FeatureServiceRegistry

KingHenne:
- [x] rebase and set base branch to `master` after #341 has been merged

### DIFF
--- a/packages/core/src/__tests__/feature-service-registry.test.ts
+++ b/packages/core/src/__tests__/feature-service-registry.test.ts
@@ -336,6 +336,30 @@ describe('FeatureServiceRegistry', () => {
       );
     });
 
+    describe('without an ExternalsValidator provided to the FeatureServiceRegistry', () => {
+      describe('with a Feature Service definition that is declaring external dependencies', () => {
+        beforeEach(() => {
+          providerDefinitionA = {
+            ...providerDefinitionA,
+            dependencies: {
+              externals: {
+                react: '^16.0.0'
+              }
+            }
+          };
+        });
+
+        it("doesn't throw an error", () => {
+          expect(() => {
+            featureServiceRegistry.registerFeatureServices(
+              [providerDefinitionA],
+              'test'
+            );
+          }).not.toThrowError();
+        });
+      });
+    });
+
     describe('with an ExternalsValidator provided to the FeatureServiceRegistry', () => {
       beforeEach(() => {
         featureServiceRegistry = new FeatureServiceRegistry({
--- a/packages/core/src/__tests__/feature-service-registry.test.ts
+++ b/packages/core/src/__tests__/feature-service-registry.test.ts
@@ -43,7 +43,7 @@ describe('FeatureServiceRegistry', () => {
   beforeEach(() => {
     mockExternalsValidator = {validate: jest.fn()};
 
-    featureServiceRegistry = new FeatureServiceRegistry(mockExternalsValidator);
+    featureServiceRegistry = new FeatureServiceRegistry();
 
     featureServiceA = {kind: 'featureServiceA'};
     bindingA = {featureService: featureServiceA};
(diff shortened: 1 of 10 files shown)


packages/core/src/feature-service-registry.ts
```diff
@@ -109,6 +109,20 @@ export interface FeatureServiceRegistryOptions {
    * registered.
    */
   readonly configs?: FeatureServiceConfigs;
+
+  /**
+   * When the {@link FeatureAppManager} is configured with a
```
clebert: If the

packages/core/src/feature-service-registry.ts
```diff
@@ -109,6 +109,20 @@ export interface FeatureServiceRegistryOptions {
...
+   * {@link FeatureAppManager#moduleLoader}, to load Feature Apps from a remote
+   * location that also provide their own Feature Services, i.e. the Feature
+   * Services are included in a different bundle than the integrator bundle, it
+   * might make sense to validate external dependencies that are required by
+   * those Feature Services against the shared dependencies that are provided by
+   * the integrator. This makes it possible that an error is already thrown when
```
clebert: This makes it possible to detect this incompatibility when registering a Feature Service with incompatible external dependencies.

packages/core/src/feature-service-registry.ts
```diff
@@ -109,6 +109,20 @@ export interface FeatureServiceRegistryOptions {
...
+   * registering a Feature Service with incompatible external dependencies. This
+   * gives the author early feedback as to whether a Feature Service is
```
clebert: -author

packages/core/src/feature-app-manager.ts
```diff
@@ -70,8 +70,28 @@ export interface FeatureAppManagerLike {
...
+   * `@feature-hub/module-loader-commonjs` package).
+   */
   readonly moduleLoader?: ModuleLoader;
+
+  /**
+   * When using a {@link #moduleLoader}, it might make sense to validate
+   * external dependencies that are required by Feature Apps against the
+   * shared dependencies that are provided by the integrator. This makes it
+   * possible that an error is already thrown when creating a Feature App with
+   * incompatible external dependencies. This gives the author early feedback as
```
clebert: -author

packages/core/src/__tests__/feature-service-registry.test.ts
```diff
@@ -339,65 +336,109 @@ describe('FeatureServiceRegistry', () => {
       );
     });
 
-    it('fails to register a Feature Service due to failing externals validation', () => {
-      const mockError = new Error('mockError');
-
-      mockExternalsValidator.validate = jest.fn(() => {
-        throw mockError;
+    describe('with an ExternalsValidator provided to the FeatureServiceRegistry', () => {
```
clebert: see missing test case, in former PR

packages/core/src/feature-service-registry.ts
KingHenne: I would prefer a similar sentence to the one we decided on for the `FeatureAppManagerOptions`:

> This makes it possible that an error is already thrown when registering a Feature Service with incompatible external dependencies, and thus enables early feedback as to whether a Feature Service is compatible with the integration environment.

packages/core/src/feature-service-registry.ts
KingHenne: fixed

packages/core/src/feature-app-manager.ts
KingHenne: fixed

packages/core/src/__tests__/feature-service-registry.test.ts
KingHenne: done

packages/core/src/feature-service-registry.ts
KingHenne: fixed

PR #344, opened 2019-02-07
docs(all): improve docs regarding to external names

clebert:


packages/core/src/externals-validator.ts
```diff
@@ -1,19 +1,19 @@
 import {satisfies, valid} from 'semver';
 
 /**
- * A map of provided node module names as keys and strict semver versions as
+ * A map of provided node external names as keys and strict semver versions as
```
fahrradflucht: I think `node` is now incorrect here

PR #351, opened 2019-02-07
test(async-ssr-manager): improve use-fake-timers util

clebert:


packages/async-ssr-manager/src/__tests__/use-fake-timers.ts
```diff
@@ -1,46 +1,45 @@
...
+    while (result.pending) {
+      await new Promise<void>(resolve => queueMacroTask(resolve));
+
       jest.runAllTicks();
       jest.runAllImmediates();
       jest.advanceTimersByTime(1);
 
       actualTimeoutInMilliseconds += 1;
+
+      await new Promise<void>(resolve => queueMicroTask(resolve));
```
KingHenne: Are the expectations off by a millisecond if this line is omitted?

packages/async-ssr-manager/src/__tests__/use-fake-timers.ts
```diff
@@ -1,46 +1,45 @@
 export type Effect<TResult> = () => Promise<TResult>;
 
-class ObservablePromise<TResult> {
+class MonitorablePromise<TResult> {
```
KingHenne: That’s a better name. 👍🏻

packages/async-ssr-manager/src/__tests__/use-fake-timers.ts
```diff
@@ -1,46 +1,45 @@
...
-  public async isPending(): Promise<boolean> {
-    for (let i = 0; i < 25; i += 1) {
-      await Promise.resolve();
-    }
-
-    return this._pending;
-  }
 }
 
+const queueMicroTask = process.nextTick.bind(process);
```
fahrradflucht: While this shouldn't make any difference here, technically the next tick queue is before micro tasks so `(f) => Promise.resolve.then(f)` should be more correct, shouldn't it?

PR #352, opened 2019-02-08
feat(react): handle feature app errors

fahrradflucht:


docs/guides/integrating-the-feature-hub.md
```diff
@@ -288,6 +288,11 @@ integrator:
 </section>
 ```
 
+If a Feature App throws an error while rendering or, in the case of a
+`ReactFeatureApp`, in a lifecycle method, the `FeatureAppContainer` will resort
+to rendering `null`. Except for on the server where rendering errors won't be
+catched and therefore have to handled by the integrator.
+
```
clebert: When a Feature App throws an error while rendering or, in the case of a `ReactFeatureApp`, throws an error in a lifecycle method, the `FeatureAppContainer` renders null. Except on the server, where rendering errors are not caught and must therefore be handled by the integrator.

packages/react/src/feature-app-container.tsx
```diff
@@ -113,22 +131,37 @@ class InternalFeatureAppContainer extends React.PureComponent<
...
+    }
   }
 }
 
 /**
  * The `FeatureAppContainer` component allows the integrator to bundle Feature
  * Apps instead of loading them from a remote location. It can also be used by
  * a Feature App to render another Feature App as a child.
+ *
+ * If a Feature App throws an error while rendering or, in the case of a {@link
```
clebert: see above

packages/react/src/feature-app-container.tsx
```diff
@@ -94,11 +103,20 @@ class InternalFeatureAppContainer extends React.PureComponent<
...
   public componentDidMount(): void {
     const container = this.containerRef.current;
 
     if (container && this.featureApp && isDomFeatureApp(this.featureApp)) {
-      this.featureApp.attachTo(container);
+      try {
+        this.featureApp.attachTo(container);
+      } catch (error) {
+        console.error(error);
+        this.setState({hasError: true});
```
clebert: maybe call `this.componentDidCatch()` instead

packages/react/src/feature-app-container.tsx
```diff
@@ -55,14 +55,23 @@ export interface FeatureAppContainerProps {
...
   typeof document === 'object' &&
   document.nodeType === 9;
 
 class InternalFeatureAppContainer extends React.PureComponent<
-  InternalFeatureAppContainerProps
+  InternalFeatureAppContainerProps,
+  InternalFeatureAppContainerState
 > {
+  public readonly state: InternalFeatureAppContainerState = {
+    hasError: false
```
clebert: Maybe rename to `hasLifecycleError` because rendering errors are handled directly in the render method.

packages/react/src/feature-app-container.tsx
```diff
@@ -113,22 +131,37 @@ class InternalFeatureAppContainer extends React.PureComponent<
   }
 
   public render(): React.ReactNode {
-    if (!this.featureApp) {
+    if (!this.featureApp || this.state.hasError) {
       return null;
```
clebert: Shouldn't we also throw the lifecycle error on the server instead of just rendering null here?

packages/react/src/feature-app-container.tsx
clebert: Maybe we should remember the complete lifecycle error instead of a boolean to re-throw it below.

packages/react/src/feature-app-container.tsx
fahrradflucht: [Error boundaries don't work on the server](https://reactjs.org/docs/error-boundaries.html) and we don't call `attachTo` there, so this won't happen there anyway.

packages/react/src/feature-app-container.tsx
fahrradflucht: See https://github.com/sinnerschrader/feature-hub/pull/352#discussion_r255296839

packages/react/src/feature-app-container.tsx
```diff
@@ -55,14 +55,23 @@ export interface FeatureAppContainerProps {
 type InternalFeatureAppContainerProps = FeatureAppContainerProps &
   FeatureHubContextValue;
 
+interface InternalFeatureAppContainerState {
+  hasFeatureAppError: boolean;
```
clebert: Please add a comment regarding the purpose of use

clebert: Will be set to true when a DOM Feature App throws in #attachTo or when the error boundary catches an error thrown by a React Feature app in a lifecycle method. Since DOM Feature Apps aren't server-side-rendered and error boundaries only work on the client, this flag will only ever be true on the client.

PR #380, opened 2019-02-25
docs(dom): add JSDoc

fahrradflucht:


packages/dom/src/feature-app-container.ts
```diff
@@ -18,6 +18,19 @@ export interface DomFeatureApp {
   attachTo(container: Element): void;
 }
 
+/**
+ * Define a custom element named `feature-app-container` at the
```
clebert: That reads like a challenge. Maybe use `Defines a ...`

packages/dom/src/feature-app-loader.ts
```diff
@@ -4,6 +4,21 @@ import {TemplateResult} from 'lit-html';
 import {until} from 'lit-html/directives/until';
 import {defineFeatureAppContainer} from './feature-app-container';
 
+/**
+ * Define a custom element named `feature-app-loader` at the
```
clebert: See the comment above.

packages/dom/src/feature-app-loader.ts
```diff
@@ -4,6 +4,21 @@ import {TemplateResult} from 'lit-html';
...
+ * The defined element has two attributes, a `src` attribute which needs to
+ * contain a URL pointing to a Feature App's module bundle and an optional
+ * `idSpecifier` attribute which needs to be defined if multiple instances of
+ * the same Feature App are placed on a single web page.
+ *
+ * It is possible to pass two slots to the `feature-app-loader` element. One
+ * slot named `loading` is rendered while the Feature App module is loading.
+ * The other one named `error` is rendered if the Feature App module could not
+ * be loaded and is also passed to the underlying `feature-app-container`
+ * element (@see {@link defineFeatureAppContainer}).
```
clebert: Isn't that an implementation detail?

fahrradflucht: Sure but by making it public to the user it is clear that the error slot behaviour of the `feature-app-container` applies as well so we don't need to duplicate it.

packages/dom/src/feature-app-container.ts
fahrradflucht: I'm pretty sure we use imperative in JSDoc everywhere (e.g. https://feature-hub.io/@feature-hub/core/classes/featureappmanager.html#getfeatureappscope)

PR #381, opened 2019-02-26
build(all): fix wallaby config (exclude dom package)

KingHenne:
On `master`, wallaby is choking on the decorators.


wallaby.js
```diff
@@ -17,12 +17,10 @@ module.exports = wallaby => ({
...
     'jest.config.js',
     'packages/*/src/**/*.{ts,tsx,snap}',
+    '!packages/{dom,demos}/**/*',
     '!packages/*/src/**/*.test.{ts,tsx}'
   ],
-  tests: [
-    'packages/*/src/**/*.test.{ts,tsx}',
-    '!packages/demos/src/**/*.test.{ts,tsx}'
-  ],
+  tests: ['packages/*/src/**/*.test.{ts,tsx}', '!packages/{dom,demos}/**/*'],
```
clebert: `!packages/{dom,demos}/**/*.{ts,tsx,snap}`

`!packages/{dom,demos}/**/*.{ts,tsx}`

KingHenne: Why?

KingHenne: Our intend is to exclude the `dom` and `demo` packages from wallaby regardless of file extensions.

PR #389, opened 2019-02-27
fix(demos): update feature-hub/react external declaration

clebert:


packages/demos/src/feature-app-in-feature-app/integrator.node.tsx
```diff
@@ -14,7 +14,7 @@ export default async function renderApp({
     moduleLoader: loadCommonJsModule,
     providedExternals: {
       react: '16.7.0',
-      '@feature-hub/react': '0.12.0'
+      '@feature-hub/react': '1.2.0'
```
KingHenne: please update `react` as well (should be `'16.8.3'`)

clebert:
@KingHenne Maybe we should take over the `getPkgVersion` function as a convenience function for integrators. It's best to integrate it somehow elegantly into `createFeatureHub` function...

packages/demos/src/get-pkg-version.ts
```diff
@@ -0,0 +1,12 @@
+import path from 'path';
+import pkgDir from 'pkg-dir';
```
KingHenne: I don’t think we need a this lib for this function. `require.resolve` should be sufficient.

packages/demos/src/server-side-rendering/feature-app.tsx
```diff
@@ -24,7 +24,8 @@ const featureAppDefinition: FeatureAppDefinition<
 
   dependencies: {
     externals: {
-      react: '^16.7.0'
+      react: '^16.7.0',
+      '@feature-hub/react': '^1.2.0'
```
KingHenne: This Feature App does not use this package, does it?

packages/demos/src/get-pkg-version.ts
clebert: nope, unfortunately not

KingHenne: Ah I see, because the resolved main file of a package can be anywhere in this package, we need to search up from there for the package.json. Correct?

clebert: yes

packages/demos/src/server-side-rendering/feature-app.tsx
clebert: i removed it

packages/demos/src/get-pkg-version.ts
KingHenne: Isn’t [find-up](
https://www.npmjs.com/package/find-up) the better fitted lib then?

clebert: It could also be used. But it doesn't matter in the end.

PR #390, opened 2019-02-28
refactor(demos): simplify getPkgVersion, remove pkg-dir dependency

KingHenne:


packages/demos/src/get-pkg-version.js
```diff
@@ -1,26 +1,13 @@
...
-      .version;
-  }
-
-  const pkgDirname = pkgDir.sync(require.resolve(pkgName));
-
-  if (!pkgDirname) {
-    throw new Error(`Unknown package '${pkgName}'.`);
-  }
+function getPkgVersion(pkgName) {
+  const pkgJsonPath = require.resolve(`${pkgName}/package.json`);
```
clebert: i dont think this will work on netlify

PR #404, opened 2019-03-06
feat(core): add logger option to FeatureAppManager, FeatureServiceRegistry, and createFeatureHub

KingHenne:
closes #402

### DIFF
--- a/packages/core/src/feature-service-registry.ts
+++ b/packages/core/src/feature-service-registry.ts
@@ -7,6 +7,7 @@ import {
   DependencyGraph,
   toposortDependencies
 } from './internal/toposort-dependencies';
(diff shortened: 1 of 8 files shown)


packages/core/src/feature-service-registry.ts
```diff
@@ -204,9 +189,13 @@ export class FeatureServiceRegistry {
 
   private readonly consumerUids = new Set<string>();
 
+  private readonly logger: Logger;
+
   public constructor(
     private readonly options: FeatureServiceRegistryOptions = {}
-  ) {}
+  ) {
+    this.logger = options.logger || console;
```
clebert: I like the following notation more: `const {logger = console} = options;`

PR #405, opened 2019-03-06
build(all): first integration of rcgen

clebert:
This PR tests a first integration of **rcgen** in the Feature Hub. Currently I have chosen Prettier as my integration target. Prettier influences configuration files of Git and VS code. In the future, everything in the `rcgen/` folder will be consumable as npm packages. Also, many of the entries in `rcgen.js` manifest file will be provided by other tool integrations. For Example:

```
excludeFilesInVscode(
  '**/.cache',
  '**/coverage', // by Jest
  '**/lerna-debug.log', // by Lerna
  '**/lib', // by typescript
  '**/node_modules', // by npm
  '**/npm-debug.log', // by npm
  '**/package-lock.json',
  '**/packages/website/build',
  '**/packages/website/i18n',
  '**/yarn-error.log' // by Yarn
),
```

It is expected that after full integration only the following remains to be configured in the `rcgen.js` manifest file:

```
excludeFilesInVscode(
  '**/.cache',
  '**/package-lock.json',
  '**/packages/website/build',
  '**/packages/website/i18n'
),
```


rcgen.js
```diff
@@ -0,0 +1,88 @@
...
+    ...prettierPatchers,
+    ...vscodePatchers,
+    ignoreFilesWithGit(
+      '.cache',
+      'coverage',
+      'lerna-debug.log',
+      'lib',
+      'node_modules',
+      'npm-debug.log',
+      'package-lock.json',
```
fahrradflucht: why would I .gitignore a lockfile?

rcgen.js
```diff
@@ -0,0 +1,88 @@
...
+      '**/yarn.lock'
+    ),
+    excludeFilesInVscode(
+      '**/.cache',
+      '**/coverage',
+      '**/lerna-debug.log',
+      '**/lib',
+      '**/node_modules',
+      '**/npm-debug.log',
+      '**/package-lock.json',
```
fahrradflucht: again I am suprised by the inconsistency here between `npm` and `yarn` lockfiles. Or is the ignorance of `package-lock.json` something that a `yarn` plugin would provide?

rcgen/git/utils.js
```diff
@@ -0,0 +1,15 @@
...
+
+const {gitIgnoreFile} = require('./files');
+const {merge} = require('@rcgen/patchers');
+
+/**
+ * @param {...string} filenames
+ *
+ * @returns {import('@rcgen/core').Patcher<string[]>}
+ */
+function ignoreFilesWithGit(...filenames) {
```
fahrradflucht: what is the reasoning for naming those functions `ignoreFilesWithFoo`? Intuitively I would name those `fooignoreFiles`...

rcgen/prettier/patchers.js
```diff
@@ -0,0 +1,20 @@
...
+const {prettierFiles} = require('./files');
+
+const prettierFilenames = prettierFiles.map(({filename}) => filename);
+
+exports.prettierPatchers = [
+  ignoreFilesWithGit(...prettierFilenames),
+  excludeFilesFromSearchInVscode(...prettierFilenames),
+  excludeFilesInVscode(...prettierFilenames),
+  recommendVscodeExtensions('esbenp.prettier-vscode'),
+  configureVscode({'editor.formatOnSave': true})
```
fahrradflucht: why is this something that prettier provides? Isn't it distinct from prettier whether or not I would want my editor to format on save?

rcgen/prettier/utils.js
```diff
@@ -0,0 +1,26 @@
...
+
+exports.configurePrettier = configurePrettier;
+
+/**
+ * @param {...string} filenames
+ *
+ * @returns {import('@rcgen/core').Patcher<string[]>}
+ */
+function ignoreFilesWithPrettier(...filenames) {
+  return merge(prettierIgnoreFile.filename, [...filenames]);
```
fahrradflucht: am I confused or is `[...filenames]` deep equal to `filenames`

rcgen.js
```diff
@@ -0,0 +1,88 @@
...
+      proseWrap: 'always',
+      singleQuote: true
+    }),
+    ignoreFilesWithPrettier(
+      '.cache',
+      'CHANGELOG.md',
+      'coverage',
+      'lib',
+      'node_modules',
+      'package.json',
```
fahrradflucht: isn't there a conceptual problem here? I imagine that you envision this line to be contributed by the `npm` plugin. But isn't this an `n`-`m` problem starting to arise? Because now all plugins for tools which write files need to know of all formatters people could use for this filetype...

.vscode/settings.json
```diff
@@ -1,20 +0,0 @@
...
-    "**/npm-debug.log": true,
-    "**/package-lock.json": true,
-    "**/packages/website/build": true,
-    "**/packages/website/i18n": true,
-    "**/yarn-error.log": true,
-    "**/yarn.lock": true
-  },
-  "tslint.enable": true,
-  "tslint.ignoreDefinitionFiles": false,
-  "typescript.tsdk": "node_modules/typescript/lib"
```
fahrradflucht: where are those setting now?

rcgen.js
```diff
@@ -0,0 +1,88 @@
...
+      'lib',
+      'node_modules',
+      'npm-debug.log',
+      'package-lock.json',
+      'packages/website/build',
+      'packages/website/i18n',
+      'todo.tasks',
+      'yarn-error.log'
+    ),
+    configurePrettier({
```
fahrradflucht: are the prettier settings typesafe here or do I lose the intellisense I would have in the regular config file provided through the public json schema?

rcgen.js
clebert: This file doesn't make sense in a Yarn Repo.

rcgen.js
clebert: This file doesn't make sense in a Yarn Repo.

rcgen/git/utils.js
clebert: Ok i will change this to `gitIgnoreFiles`

rcgen/prettier/utils.js
clebert: shame on me

.vscode/settings.json
clebert: doesn't need anymore

rcgen.js
clebert: i will try to combine json-schema and typescript for this issue

package.json
```diff
@@ -10,6 +10,7 @@
     "compile": "lerna exec -- tsc --project tsconfig.cjs.json",
     "compile:esm": "lerna exec --no-private -- tsc --project tsconfig.esm.json",
     "format": "prettier --write '**/*.{js,json,md,ts,tsx,yml}'",
+    "postinstall": "ts-node node_modules/@rcgen/cli/bin/run --verbose --force",
```
fahrradflucht: You could evaluate using `ts-node/register` or even something like [`interpret`](https://www.npmjs.com/package/interpret) to support typescript configs in general...

package.json
```diff
@@ -36,6 +37,11 @@
     "@commitlint/cli": "^7.2.1",
     "@commitlint/config-angular": "^7.1.2",
     "@commitlint/travis-cli": "^7.2.1",
+    "@feature-hub/rcgen": "^0.0.0",
+    "@rcgen/cli": "^0.8.0",
+    "@rcgen/core": "^0.8.0",
+    "@rcgen/filetypes": "^0.8.0",
+    "@rcgen/patchers": "^0.8.0",
```
fahrradflucht: Do we really need all of those deps here and in the package?

packages/rcgen/src/format.ts
```diff
@@ -0,0 +1,35 @@
...
+  } = options;
+
+  return composeEnhancers([
+    enhancePrettierConfig({
+      useTabs:
+        indentStyle === 'tab'
+          ? true
+          : indentStyle === 'space'
+          ? false
+          : undefined,
```
fahrradflucht: What is the difference between setting `useTabs` to `undefined` vs `false`?

clebert: With undefined you have a fallback to the default prettier behavior, because undefined will not be serialized to json

package.json
clebert: yes because of the linters, and they are our dev dependencies, so in my opinion it is correct

package.json
clebert: since this is only a temporary solution, which will disappear again soon, but works now, i don't want to put much effort into it

PR #408, opened 2019-03-08
feat(react): add custom logger option to FeatureHubContext

KingHenne:
The `FeatureHubContext` provides the custom logger to the `FeatureAppLoader` and `FeatureAppContainer`.

### DIFF
--- a/packages/react/src/feature-hub-context.tsx
+++ b/packages/react/src/feature-hub-context.tsx
@@ -1,18 +1,26 @@
 import {AsyncSsrManagerV1} from '@feature-hub/async-ssr-manager';
-import {FeatureAppManager} from '@feature-hub/core';
+import {FeatureAppManager, Logger} from '@feature-hub/core';
 import * as React from 'react';
 
-export interface FeatureHubContextValue {
+type SomeRequired<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>> &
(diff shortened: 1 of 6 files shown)


packages/react/src/feature-hub-context.tsx
```diff
@@ -26,13 +34,20 @@ export interface FeatureHubContextValue {
   addUrlForHydration?(url: string): void;
 }
 
+export type FeatureHubContextValue = FeatureHubContextProviderValue;
```
KingHenne: I will add the following deprecating warning in a separate `fix` commit, where I also need to replace the deprecated interface in `packages/demos/src/server-side-rendering/integrator.node.tsx`.

```ts
/**
 * @deprecated Use {@link FeatureHubContextProviderValue} or
 * {@link FeatureHubContextConsumerValue} instead.
 */
```

Otherwise `@feature-hub/demos` would also get a minor version bump, and the feature would be mentioned in its changelog.

KingHenne: If we would allow fast-forward merges, I could even add the commit to this PR. 😉

clebert: Do as discussed.

KingHenne: https://github.com/sinnerschrader/feature-hub/pull/411

PR #415, opened 2019-03-12
feat(async-ssr-manager): use Logger Feature Service instead of console

unstubbable:
The integrator can now provide the Logger Feature Service, which the Async SSR Manager then uses for logging. The Async SSR Manager declares the Logger Feature Service as an optional dependency, and uses `console` as a fallback logger, when the Logger Feature Service is not provided.

### DIFF
--- a/packages/async-ssr-manager/src/index.ts
+++ b/packages/async-ssr-manager/src/index.ts
@@ -87,7 +87,7 @@ export interface SharedAsyncSsrManager extends SharedFeatureService {
 }
 
 export interface AsyncSsrManagerDependencies extends FeatureServices {
-  's2:logger'?: Logger;
+  readonly 's2:logger'?: Logger;
 }
 
 /**
--- a/packages/async-ssr-manager/src/index.ts
+++ b/packages/async-ssr-manager/src/index.ts
@@ -1,9 +1,12 @@
 import {
   FeatureServiceBinder,
   FeatureServiceProviderDefinition,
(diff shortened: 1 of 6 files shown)


packages/async-ssr-manager/src/internal/async-ssr-manager-context.ts
```diff
@@ -0,0 +1,14 @@
+import {Logger} from '@feature-hub/logger';
+import {AsyncSsrManagerDependencies} from '..';
+
+export interface AsyncSsrManagerContext {
+  logger: Logger;
```
clebert: readonly

clebert: please fix also everywhere else

packages/async-ssr-manager/src/index.ts
```diff
@@ -83,19 +86,32 @@ export interface SharedAsyncSsrManager extends SharedFeatureService {
   readonly '1.0.0': FeatureServiceBinder<AsyncSsrManagerV1>;
 }
 
+export interface AsyncSsrManagerDependencies extends FeatureServices {
+  's2:logger'?: Logger;
```
clebert: readonly

clebert: please fix it also in history

unstubbable: see #416

PR #416, opened 2019-03-12
fix(history-service): add missing readonlys

unstubbable:


packages/history-service/src/define-history-service.ts
```diff
@@ -29,8 +29,8 @@ export interface SharedHistoryService extends SharedFeatureService {
 }
```
clebert: add readonly to line 21

PR #399, opened 2019-03-01
docs(website): improve Feature Hub introduction

clebert:
Closes #332

PR #421, opened 2019-03-13
remove examples

chaosmonster:
It is not working with at least two of the mentioned framework/ technologies. For that reason please remove this, so no one will try and fail

### DIFF
--- a/docs/getting-started/introduction.md
+++ b/docs/getting-started/introduction.md
@@ -26,7 +26,7 @@ provide a full-fledged solution for composing micro frontends:
   [placing micro frontends on a web page using
   React][placing-feature-apps-on-a-web-page-using-react], but also allows the
   integration of micro frontends that are [built with any other frontend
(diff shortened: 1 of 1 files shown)


unstubbable:
Could you please open an issue instead, and describe what you tried to accomplish and what exactly failed? This section should still be valid, although it will be updated soon in #399.

clebert:
I do not find such an unfounded statement very constructive, let alone truthful.

I therefore felt compelled to set up an example repository showing the integration of an Angular Feature App.

It took me about 3 hours to build the example, but I would like to point out that I have never developed an Angular application before. I think an experienced Angular developer would only need a fraction of the time.

Here is the link to the repo: https://github.com/clebert/angular-feature-app

@fahrradflucht @unstubbable I think we should also create an example integration with Vue.js

chaosmonster:
Thanks for the example. I will withdraw this PR and be happy to also see a vue.js example

PR #429, opened 2019-03-15
build(all): setup GitHub actions

clebert:


clebert:
Closed due to poor performance. We will change from Travis to Azure.

PR #435, opened 2019-03-17
build(all): update rcgen to v0.13.0

clebert:


rcgen.js
```diff
@@ -1,23 +1,30 @@
...
+    'lib',
+    'node_modules',
+    'npm-debug.log',
+    'package-lock.json',
+    'packages/website/build',
+    'packages/website/i18n',
+    'todo.tasks',
+    'yarn-error.log'
+  ),
+  gitIgnoreIntrinsicFiles({excludedFilenamePatterns: ['package.json']}),
```
unstubbable: I don't understand why the `package.json` would be ignored by git "intrinsically".

clebert: Because `npm()` is included two lines below and defines the `package.json` file. Currently the `package.json` file is already generated. This means read in but not yet modified, because there is no config provider there who wants to change something about it.

clebert: Perhaps the term "intrinsic" is inappropriate? I found it better than e.g. "internal".

PR #448, opened 2019-03-28
fix(module-loader-commonjs): avoid require cache

unstubbable:
While server-side rendering Feature Apps in development mode, no changes to the Feature App code are visible in the rendered html, since the Feature App server bundle is always loaded from the require cache while the server is running. This fix ensures that no stale version of a fetched module is loaded from the require cache.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .cache
 .node-version
 .nvmrc
+.tmp
 coverage
 lerna-debug.log
 lib
--- a/packages/module-loader-commonjs/src/index.ts
+++ b/packages/module-loader-commonjs/src/index.ts
@@ -5,7 +5,7 @@ import fetch from 'node-fetch';
 import path from 'path';
 
 // This must be a local directory to make node module resolution possible.
-const cacheDirname = path.join(__dirname, '.cache');
+const tmpDirname = path.join(__dirname, '.tmp');
 
 function createHash(data: string): string {
   return crypto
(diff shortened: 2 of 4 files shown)


unstubbable:
I tried to write a unit test for this, but Jest [does not implement require.cache](https://github.com/facebook/jest/issues/5120).

packages/module-loader-commonjs/src/index.ts
```diff
@@ -5,7 +5,7 @@ import fetch from 'node-fetch';
 import path from 'path';
 
 // This must be a local directory to make node module resolution possible.
-const cacheDirname = path.join(__dirname, '.cache');
+const tmpDirname = path.join(__dirname, '.tmp');
```
clebert: Why do you changed the name to tmp?

.gitignore
```diff
@@ -1,6 +1,7 @@
 .cache
```
clebert: is .cache needed?

packages/module-loader-commonjs/src/index.ts
unstubbable: Because `cache` is misleading, since we explicitly do not cache fetched modules (even before this fix). They are only written to disk to make them `require`able.

.gitignore
unstubbable: I left it in to not create untracked files for other developers that already have a `.cache` folder.

PR #452, opened 2019-03-29
feat(react): support adding Feature App stylesheets to the document during SSR

unstubbable:
With this feature, the integrator can prevent flickering when a server-rendered Feature App with external stylesheets is hydrated in the browser.

Normally, in the client, the `FeatureAppLoader` appends the stylesheets after it has been mounted. Since the integrator usually preloads Feature App client bundles before starting the hydration, the server-rendered Feature App HTML is displayed unstyled until the integrator client bundle as well as all Feature App client bundles have been loaded and evaluated.

Now, the integrator can collect all stylesheets during SSR, and append link elements for them to the document head, before sending the HTML to the client. This allows the browser to immediately load theses stylesheets before rendering the DOM, and thus prevents any flickering caused by missing styles.

Thanks for reporting this issue, @elbkind!

[Docs](https://deploy-preview-452--feature-hub.netlify.com/docs/guides/server-side-rendering#adding-stylesheets-to-the-document)

### DIFF
--- a/packages/demos/src/start-server.ts
+++ b/packages/demos/src/start-server.ts
@@ -1,3 +1,4 @@
+import {Css} from '@feature-hub/react';
 import express from 'express';
 import getPort from 'get-port';
 import {Server} from 'http';
@@ -12,6 +13,7 @@ export interface AppRendererOptions {
 export interface AppRendererResult {
   html: string;
   serializedStates?: string;
+  stylesheetsForSsr?: Map<string, Css>;
   urlsForHydration?: Set<string>;
 }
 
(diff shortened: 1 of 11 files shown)


packages/demos/src/start-server.ts
```diff
@@ -12,18 +13,34 @@ export interface AppRendererOptions {
...
 
 export type AppRenderer = (
   options: AppRendererOptions
 ) => Promise<AppRendererResult>;
 
+function createStylesheetLink({href, media = 'all'}: Css): string {
+  return `<link href="${href}" media="${media}" rel="stylesheet" />`;
+}
+
+function createStylesheetLinks(stylesheets: Css[]): string {
```
clebert: Maybe this can be changed to `stylesheetsForSsr: Set<string> | undefined`

packages/demos/src/start-server.ts
```diff
@@ -12,18 +13,34 @@ export interface AppRendererOptions {
...
-  serializedStates?: string,
-  urlsForHydration?: Set<string>
+  {
+    serializedStates,
+    stylesheetsForSsr,
+    urlsForHydration
+  }: Partial<AppRendererResult> = {}
 ): string {
+  const stylesheetLinks = stylesheetsForSsr
+    ? createStylesheetLinks(Array.from(stylesheetsForSsr.values()))
```
clebert: with the above note the call could be simplified and the conditional removed

unstubbable: I thought about this, but I kind of like the symmetry with `serializedStatesScript` and `urlsForHydrationScript` below. Furthermore, I'm not so fond (anymore) of helper functions that receive a single parameter, that is then made optional and has the undefined guard inside, instead of just not calling this function from the outside.

unstubbable: 🤔 Or, we could of course go the other way and also create functions for the other two cases.

clebert: It's ok for me.

PR #464, opened 2019-04-03
build(all): establish compatibility with the latest version of TSLint

clebert:


packages/core/src/__tests__/feature-app-manager.test.ts
```diff
@@ -94,6 +94,7 @@ describe('FeatureAppManager', () => {
       expect(asyncFeatureAppDefinition.error).toBeUndefined();
     });
 
+    // tslint:disable-next-line no-null-undefined-union
```
unstubbable: Where did this issue arise?

clebert: build(all): lock file maintenance

clebert: But since the rule also complains about types like ReactNode, I deactivated it completely.

unstubbable: I guess this line can be removed since you disabled it in the config file.

PR #476, opened 2019-04-15
feat(all): specify a Feature App's base URL

unstubbable:
This allows an integrator to specify a relative or absolute `baseUrl` for two purposes:

1.) as a common base URL for the `src`, `serverSrc`, and `css` hrefs
2.) to provide the Feature App with a base URL that it can use to reference its own assets and/or BFF


packages/react/src/feature-app-loader.tsx
```diff
@@ -6,20 +6,30 @@ import {
...
-   * The URL of the Feature App's server module bundle.
+   * The URL of the Feature App's server module bundle. If [[baseUrl]] is
+   * specified, it will be prepended, unless `serverSrc` is an absolute URL.
+   * Either [[baseUrl]] or `serverSrc` must be an absolute URL.
    */
   readonly serverSrc?: string;
 
   /**
-   * A list of stylesheets that should be added to the document.
+   * A list of stylesheets that should be added to the document. If [[baseUrl]]
```
janstoff: Suggestion: simply `to be added to` instead of `that should be added to`.

packages/react/src/internal/prepend-base-url.ts
```diff
@@ -0,0 +1,17 @@
+import urljoin from 'url-join';
+
+function isAbsolute(url: string): boolean {
+  return /^https?:\/\//.test(url);
+}
+
+/**
+ * Prepends the given `url` with the given `baseUrl` (can be relative or
```
janstoff: Suggestion: "Prepends the given `baseURL` - which can be relative or absolute - to the `url` , unless the `url` is absolute".

PR #500, opened 2019-06-05
feat(core): remove ability to provide configs via FeatureServiceRegistry

unstubbable:
BREAKING CHANGE: The option `featureServiceConfigs` has been removed from the options of `createFeatureHub` and from the options of the `FeatureServiceRegistry` constructor. The `env` that is passed to a Feature Service's `create` method does not include a `config` property anymore. If a Feature Service must be configured, a factory function that accepts options, and that returns a Feature Service definition, should be used instead, see `@feature-hub/async-ssr-manager` for an example.

---

***Note:** The `primaryConsumerUid` option of the History Service will be renamed to `primaryConsumerId` in a separate PR, since this would change the API of the `@feature-hub/history-service` package, whereas this PR only changes the API of `@feature-hub/core`.*

### DIFF
--- a/packages/core/src/feature-app-manager.ts
+++ b/packages/core/src/feature-app-manager.ts
@@ -96,7 +96,7 @@ export interface FeatureAppScopeOptions {
    * A callback that is called before the Feature App is created.
    */
   readonly beforeCreate?: (
-    consumerId: string,
+    featureAppUid: string,
     featureServices: FeatureServices
   ) => void;
 }
--- a/packages/core/src/feature-app-manager.ts
+++ b/packages/core/src/feature-app-manager.ts
@@ -52,6 +52,8 @@ export interface FeatureAppDefinition<
   TInstanceConfig = unknown,
   TFeatureServices extends FeatureServices = FeatureServices
 > extends FeatureServiceConsumerDefinition {
+  readonly id: string;
+
   readonly ownFeatureServiceDefinitions?: FeatureServiceProviderDefinition<
     SharedFeatureService
   >[];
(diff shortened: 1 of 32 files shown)


packages/demos/src/custom-logging/index.test.ts
```diff
@@ -45,13 +45,13 @@ describe('integration test: "custom logging"', () => {
     expect(messages).toContainEqual([
       '%ctest:integrator',
       'font-weight: bold',
-      'Creating Feature App with consumerUid "test:logging-app:first"...'
```
clebert: Please rename consumerId to ID

packages/demos/src/custom-logging/integrator.tsx
```diff
@@ -36,10 +36,8 @@ const logger = featureServices['s2:logger'] as Logger;
 ReactDOM.hydrate(
   <FeatureHubContextProvider value={{featureAppManager}}>
     <App
-      beforeCreate={consumerUid =>
-        logger.debug(
-          `Creating Feature App with consumerUid "${consumerUid}"...`
-        )
+      beforeCreate={consumerId =>
+        logger.debug(`Creating Feature App with consumerId "${consumerId}"...`)
```
clebert: Please rename consumerId to ID

packages/core/src/feature-app-manager.ts
```diff
@@ -94,7 +96,7 @@ export interface FeatureAppScopeOptions {
    * A callback that is called before the Feature App is created.
    */
   readonly beforeCreate?: (
-    consumerUid: string,
+    consumerId: string,
```
clebert: rename to featureAppId

PR #504, opened 2019-06-07
feat(all): let featureAppId be defined by the integrator

### LINKED ISSUE #495: ConsumerUid of feature apps
The consumerUid of feature apps is currently constructed from the feature app definition id and an id specifier. The main problem with this approach is, that a feature app is hard to reference by the integrator. The integrator defines the idSpecifier and the javascript url of the feature app. To derive the consumerUid from feature app definition the integrator must load the javascript  from the url.

I suggest to not prepend the feature app definition id to the idSpecifier if it is defined. This way it's the  integrator's responsibility to make sure the consumerUid is unique and it can manage references to feature apps.

unstubbable:
fixes #495

BREAKING CHANGE: The option `featureAppConfigs` has been removed from the options of `createFeatureHub` and from the options of the `FeatureAppManager` constructor. The `env` that is passed to a Feature App's `create` method does not include a `config` property anymore. If a Feature App must be configured, the integrator needs to specify the `config` prop of the `FeatureAppLoader` or `FeatureAppContainer`. Furthermore, the `FeatureAppLoader` or `FeatureAppContainer` now require a `featureAppId` prop, and the `instanceConfig` and `idSpecifier` props have been removed. The same applies to the `<feature-app-loader>` and `<feature-app-container>` custom elements. Since the integrator now needs to define the ID of a Feature App, the Feature App definition must not specify an `id` anymore.

### DIFF
--- a/packages/react/src/feature-app-container.tsx
+++ b/packages/react/src/feature-app-container.tsx
@@ -44,6 +44,15 @@ export type FeatureApp = ReactFeatureApp | DomFeatureApp;
 
 export interface FeatureAppContainerProps {
   /**
+   * The Feature App ID is used to identify the Feature App instance. Multiple
+   * Feature App Loaders with the same `featureAppId` will render the same
+   * Feature app instance. The ID is also used as a consumer ID for dependent
+   * Feature Services. To render multiple instances of the same kind of Feature
+   * App, different IDs must be used.
+   */
+  readonly featureAppId: string;
+
+  /**
    * The absolute or relative base URL of the Feature App's assets and/or BFF.
    */
   readonly baseUrl?: string;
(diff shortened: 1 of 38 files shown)


packages/react/src/feature-app-container.tsx
```diff
@@ -104,16 +105,17 @@ class InternalFeatureAppContainer extends React.PureComponent<
...
-      instanceConfig
+      featureAppId,
+      featureAppManager
     } = props;
 
     try {
       this.featureAppScope = featureAppManager.getFeatureAppScope(
         featureAppDefinition,
-        {baseUrl, idSpecifier, instanceConfig, beforeCreate}
+        featureAppId,
```
clebert: please change the sort order of the first/second args

packages/react/src/feature-app-loader.tsx
```diff
@@ -192,10 +193,10 @@ class InternalFeatureAppLoader extends React.PureComponent<
     return (
       <FeatureAppContainer
         baseUrl={baseUrl}
-        featureAppDefinition={featureAppDefinition}
-        idSpecifier={idSpecifier}
-        instanceConfig={instanceConfig}
         beforeCreate={beforeCreate}
+        config={config}
+        featureAppDefinition={featureAppDefinition}
```
clebert: Please change order as above

PR #506, opened 2019-06-10
feat(all): pass full env to beforeCreate callback

### LINKED ISSUE #490: Proposal: Breaking change of beforeCreate
Maybe it would be better to pass the whole environment to `beforeCreate` instead of just the bound feature services. The environment is also the argument for the `create` method and contains potentially useful information.

New:

```ts
readonly beforeCreate?: (
  consumerUid: string,
  env: FeatureAppEnvironment<TConfig, TInstanceConfig, TFeatureServices>
) => void;
```

Current:

```ts
readonly beforeCreate?: (
  consumerUid: string,
  featureServices: FeatureServices
) => void;
```

unstubbable:
fixes #490

BREAKING CHANGE: Instead of passing the `featureAppId` and the `featureServices` as separate arguments to the `beforeCreate` callback, the full Feature App environment, that contains the `featureAppId` and `featureServices`, is now passed as a single argument (`env`). This is the same argument that is passed to the Feature App's `create` method.

### DIFF
--- a/packages/core/src/feature-app-manager.ts
+++ b/packages/core/src/feature-app-manager.ts
@@ -23,7 +23,7 @@ export interface FeatureAppEnvironment<
   /**
    * A config object that is provided by the integrator.
    */
-  readonly config: TConfig;
+  readonly config: TConfig | undefined;
 
   /**
    * The ID that the integrator has assigned to the Feature App instance.
@@ -56,7 +56,7 @@ export interface FeatureAppScope<TFeatureApp> {
   destroy(): void;
 }
 
-export interface FeatureAppScopeOptions {
+export interface FeatureAppScopeOptions<TConfig> {
   /**
    * The absolute or relative base URL of the Feature App's assets and/or BFF.
    */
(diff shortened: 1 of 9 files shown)


packages/core/src/feature-app-manager.ts
```diff
@@ -56,7 +56,7 @@ export interface FeatureAppScope<TFeatureApp> {
   destroy(): void;
 }
 
-export interface FeatureAppScopeOptions {
+export interface FeatureAppScopeOptions<TConfig> {
```
clebert: Maybe it makes sense that the FeatureServices can also be overwritten? Useful for typed access.

unstubbable: For `FeatureAppScopeOptions` which is only used (internally more or less) by the `FeatuerAppContainer` this does not really matter. You could argue though that this might make sense for `FeatureAppContainer`/`FeatureAppContainerProps`. But in contrast to `TConfig` where the type of the `config` prop matches the config property of `env`, the type of the Feature Services in `env` has no counterpart in `FeatureAppContainerProps`. Therefore a type param wouldn't be any better than a type cast. See also https://twitter.com/SeaRyanC/status/1118634571564630016

unstubbable: Nevermind, see c01c1f6.

PR #508, opened 2019-06-11
feat(react): do not re-throw render errors on server

unstubbable:
BREAKING CHANGE: When the FeatureAppContainer/FeatureAppLoader catches a render error on the server, this error was previously re-thrown. To re-create this behavior, an `onError` prop must be defined that throws the error.


packages/react/src/__tests__/feature-app-container.node.test.tsx
```diff
@@ -21,14 +21,6 @@ describe('FeatureAppContainer (on Node.js)', () => {
   let mockFeatureAppScope: FeatureAppScope<unknown>;
```
clebert: Shouldn't we explicitly test the rethrow with a method that has been passed in?

PR #514, opened 2019-06-17
docs(all): id and config changes

unstubbable:


docs/guides/writing-a-feature-app.md
```diff
@@ -4,14 +4,12 @@ title: Writing a Feature App
 sidebar_label: Writing a Feature App
 ---
 
-A Feature App is described by a consumer definition object. It consists of an
-`id`, `dependencies` and/or `optionalDependencies` objects, and a `create`
+A Feature App is described by a consumer definition object. It consists of
+optional `dependencies` and `optionalDependencies` objects, and a `create`
```
clebert: maybe write `both optional ...`

docs/guides/writing-a-feature-service.md
```diff
@@ -150,7 +134,7 @@ const myFeatureServiceDefinition = {
   id: 'acme:my-feature-service',
 
   create(env) {
-    let count = env.config.initialCount || 0;
+    let count = 0;
```
clebert: Shouldn't you continue to use a config here and show how to do it?

unstubbable: No, I want to avoid the noise here. We have a separate doc for that.

PR #521, opened 2019-07-05
test(all): remove jest-stub-methods

unstubbable:
With this change we can avoid the errors in #494.


package.json
```diff
@@ -15,9 +15,9 @@
     "lint": "lerna exec --parallel -- tslint --config tslint.json --project tsconfig.json --format verbose",
     "sort-package-jsons": "sort-package-json package.json packages/*/package.json",
     "pretest": "rimraf coverage",
-    "test": "jest --verbose",
-    "test:integration": "jest --verbose --testPathPattern packages/demos",
-    "test:unit": "jest --verbose --testPathIgnorePatterns packages/demos",
+    "test": "jest --verbose --silent",
```
clebert: The --silent does not seem to work.

unstubbable: It does work for the test output, but not for the webpack dev server output. Looking into it...

unstubbable: Scratch that, it also works for the integration tests with webpack, I just forgot to also set it in the ci script (done in 17cdc0d).

PR #533, opened 2019-07-27
fix(module-loader-commonjs): avoid using fs to require commonjs module

### LINKED ISSUE #503: module-loader-commonjs prevents running on read only filesystem
When running on a read only filesystem like aws lambda you get a write error.
In module-loader-commonjs/index.ts line 20 a mkdirp fails.

unstubbable:
fixes #503


unstubbable:
@fahrradflucht There was no cache before. The file was always fetched and newly written into tmp.

PR #539, opened 2019-08-26
feat(history-service): root location transformer createNewRootLocationForMultipleConsumers

unstubbable:
FYI @stemey


packages/history-service/src/internal/history-multiplexer.ts
```diff
@@ -108,4 +117,17 @@ export class HistoryMultiplexer {
       state: newConsumerStates
     });
   }
+
+  private setConsumerStates(
```
clebert: This method is more of a helper function and does not require access to the `this` object. So wouldn't a function be better suited?

packages/history-service/src/internal/history-multiplexer.ts
```diff
@@ -72,17 +73,25 @@ export class HistoryMultiplexer {
...
+          consumerLocation.historyKey
+        );
+      }
     }
 
-    return rootlocation;
+    return this.setConsumerStates(newRootLocation, ...consumerLocations);
   }
 
   private createRootLocation(
```
clebert: Shouldn't we also use `setConsumerStates` in this function?

unstubbable: no, here we only add a single consumer state instead of setting/replacing all of them at once

packages/history-service/src/internal/history-multiplexer.ts
unstubbable: done

PR #567, opened 2020-03-19
feat(module-loader-commonjs): add proxy support

niels-garve:
Dear Feature Hub team,

We at Das Büro am Draht need to load Feature Apps via proxy.

The dependency `node-fetch` does not consider proxy settings that are defined by environment variable (`process.env.HTTP_PROXY`).

Also, it seems that the team around `node-fetch` won't ever add proxy support. Please see this issue for further information: https://github.com/node-fetch/node-fetch/issues/195

So, with this pull request I add proxy support to the module loader by adding a proxy agent to the request.

What do you think? I hope you like the feature and its implementation.

Looking forward to hearing from you.

Thanks in advance and kind regards
Niels Garve (Das Büro am Draht)

### DIFF
--- a/packages/module-loader-commonjs/src/__tests__/index.test.ts
+++ b/packages/module-loader-commonjs/src/__tests__/index.test.ts
@@ -2,16 +2,22 @@
  * @jest-environment node
  */
 
+import {RequestInit} from 'node-fetch';
 import {createCommonJsModuleLoader, loadCommonJsModule} from '..';
 
 let mockResponse: string;
+let requestUrl: RequestInfo;
+let requestInit: RequestInit;
 
 // tslint:disable promise-function-async
-jest.mock('node-fetch', () => () =>
-  Promise.resolve({
+jest.mock('node-fetch', () => (url: RequestInfo, init: RequestInit) => {
+  requestUrl = url;
+  requestInit = init;
+
+  return Promise.resolve({
     text: () => Promise.resolve(Buffer.from(mockResponse))
-  })
-);
+  });
+});
 // tslint:enable promise-function-async
 
 describe('loadCommonJsModule (on Node.js)', () => {
@@ -35,6 +41,9 @@ describe('loadCommonJsModule (on Node.js)', () => {
 describe('createCommonJsModuleLoader', () => {
   it('creates a CommonJS module loader with custom-defined externals', async () => {
     const url = 'http://example.com/test.js';
+    const init: RequestInit = {
+      method: 'post'
+    };
 
     mockResponse = `
 			var foo = require('foo');
@@ -43,13 +52,18 @@ describe('createCommonJsModuleLoader', () => {
 			};
 		`;
 
-    const loadCommonJsModuleWithExternals = createCommonJsModuleLoader({
-      foo: () => 42
-    });
+    const loadCommonJsModuleWithExternals = createCommonJsModuleLoader(
+      {
+        foo: () => 42
+      },
+      init
+    );
 
     const loadedModule = await loadCommonJsModuleWithExternals(url);
     const expectedModule = {default: {test: 42}};
 
     expect(loadedModule).toEqual(expectedModule);
+    expect(requestUrl).toEqual(url);
+    expect(requestInit).toEqual(init);
   });
 });
--- a/packages/module-loader-commonjs/src/index.ts
+++ b/packages/module-loader-commonjs/src/index.ts
@@ -1,15 +1,16 @@
 import {ModuleLoader} from '@feature-hub/core';
-import fetch from 'node-fetch';
+import fetch, {RequestInit} from 'node-fetch';
 
 export interface Externals {
   readonly [externalName: string]: unknown;
 }
 
 export function createCommonJsModuleLoader(
-  externals: Externals = {}
+  externals: Externals = {},
+  requestInit?: RequestInit
 ): ModuleLoader {
   return async (url: string): Promise<unknown> => {
-    const response = await fetch(url);
+    const response = await fetch(url, requestInit);
     const source = await response.text();
     const mod = {exports: {}};
 


clebert:
Hello Niels, thanks for your PR. In my view, there's nothing to be said against adding this functionality. We will do a code review and give you feedback as soon as possible.

niels-garve:
> Hello Niels, thanks for your PR. In my view, there's nothing to be said against adding this functionality. We will do a code review and give you feedback as soon as possible.

Hi @clebert , thanks for your reply. Sounds amazing! Meanwhile I fixed the issues reported by the pipeline.

packages/module-loader-commonjs/src/__tests__/index.test.ts
```diff
@@ -2,16 +2,23 @@
  * @jest-environment node
  */
 
+import {RequestInfo, RequestInit} from 'node-fetch';
+import ProxyAgent from 'proxy-agent';
 import {createCommonJsModuleLoader, loadCommonJsModule} from '..';
 
 let mockResponse: string;
+let requestUrl: RequestInfo;
+let requestInitOptions: RequestInit;
```
clebert: Pleas rename to `requestInit`

packages/module-loader-commonjs/src/__tests__/index.test.ts
```diff
@@ -2,16 +2,23 @@
...
 
 // tslint:disable promise-function-async
-jest.mock('node-fetch', () => () =>
-  Promise.resolve({
+jest.mock('node-fetch', () => (url: RequestInfo, options: RequestInit) => {
```
clebert: Please change parameter names to `(url: RequestInfo, init: RequestInit) =>`

packages/module-loader-commonjs/src/__tests__/index.test.ts
```diff
@@ -52,4 +59,30 @@ describe('createCommonJsModuleLoader', () => {
 
     expect(loadedModule).toEqual(expectedModule);
   });
+
+  it('sets agent if process.env.HTTP_PROXY is defined', async () => {
+    const currentHttpProxy = process.env.HTTP_PROXY;
+    const url = 'http://example.com/test.js';
+
+    process.env.HTTP_PROXY = 'http://';
```
clebert: We have an undocumented newline rule in this project. Please add a newline between line 67 and 68 :)

packages/module-loader-commonjs/src/__tests__/index.test.ts
```diff
@@ -52,4 +59,30 @@ describe('createCommonJsModuleLoader', () => {
 
     expect(loadedModule).toEqual(expectedModule);
   });
+
+  it('sets agent if process.env.HTTP_PROXY is defined', async () => {
+    const currentHttpProxy = process.env.HTTP_PROXY;
```
clebert: I would prefer the name `initialHttpProxy`

packages/module-loader-commonjs/src/index.ts
```diff
@@ -9,7 +11,15 @@ export function createCommonJsModuleLoader(
   externals: Externals = {}
 ): ModuleLoader {
   return async (url: string): Promise<unknown> => {
-    const response = await fetch(url);
+    const requestInit: RequestInit = {};
+
+    if (process.env.HTTP_PROXY) {
+      requestInit.agent = (new ProxyAgent(
+        process.env.HTTP_PROXY
+      ) as unknown) as Agent;
```
unstubbable: This type cast caught my attention. Apparently only the latest version of `agent-base` matches the `Agent` interface of Node. This version is [not yet used in `proxy-agent`](https://github.com/TooTallNate/node-proxy-agent/issues/51). But it is used in `http-proxy-agent` as well as `https-proxy-agent`. Which made me think why you called the env `HTTP_PROXY`. Do you only need http, or should this also cover https (probably)? ... Does the feature hub even need to handle all of this? We already have a factory function for the CommonJS module loader, so why not pass the agent as an optional second param here? Then you can use whatever proxy lib you need in the integrator (still based on an env variable if you like). Maybe we should even allow the whole `requestInit` object to be passed instead of only the `agent`? What do you think, @clebert?

clebert: Yeah, that's a good idea. This allows for greater flexibility on the integrator side. So I would be in favor of passing the whole requestInit object in.

packages/module-loader-commonjs/src/__tests__/index.test.ts
niels-garve: Hi @clebert , agreed, I changed, thanks!

packages/module-loader-commonjs/src/__tests__/index.test.ts
niels-garve: @clebert ja, better, done!

packages/module-loader-commonjs/src/__tests__/index.test.ts
niels-garve: Hi @clebert I fixed it :)

packages/module-loader-commonjs/src/__tests__/index.test.ts
niels-garve: Hi @clebert, okay, I fixed it, thanks!

packages/module-loader-commonjs/src/index.ts
niels-garve: Hi @unstubbable ,
Hi @clebert ,

Thanks for your feedback!

So, creating a feature hub on our side would look like:

```javascript
const requestInit: RequestInit = {};

if (process.env.HTTP_PROXY) {
  requestInit.agent = (new ProxyAgent(
    process.env.HTTP_PROXY
  ) as unknown) as Agent;
}

createFeatureHub('integrator-id', {
  // ...
  moduleLoader: createCommonJsModuleLoader(requestInit),
  // ...
});
```

Right?

How would you like to pass `requestInit` to `createCommonJsModuleLoader`? I'd prefer an options object since it is extendable in the future, like:
```javascript
export interface Options {
  externals: Externals,
  requestInit: RequestInit
}

export function createCommonJsModuleLoader(options: Options): ModuleLoader
```

But this is a breaking change. We could also add a second parameter?

It's up to you :)

Kind regards
Niels

packages/module-loader-commonjs/src/__tests__/index.test.ts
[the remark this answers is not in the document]
niels-garve: Hi @unstubbable ,
Thanks for your feedback. Imho it's better not to change global settings from within a test to avoid side effects. For now, there is no side effect, of course. Still feels better to me. What do you think?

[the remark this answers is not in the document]
unstubbable: This will be gone anyways with the other change.

packages/module-loader-commonjs/src/index.ts
unstubbable: As I said, use a second param so we are not breaking.

packages/module-loader-commonjs/src/__tests__/index.test.ts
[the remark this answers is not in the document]
niels-garve: Hi @unstubbable the tests are removed

packages/module-loader-commonjs/src/index.ts
niels-garve: Hi @unstubbable , Hi @clebert ,
I added the second optional parameter. I squashed the commits and changed the commit message since the scope changed.

packages/module-loader-commonjs/src/index.ts
```diff
@@ -1,15 +1,16 @@
...
 }
 
 export function createCommonJsModuleLoader(
-  externals: Externals = {}
+  externals: Externals = {},
+  requestInit?: RequestInit
 ): ModuleLoader {
   return async (url: string): Promise<unknown> => {
-    const response = await fetch(url);
+    const response = await fetch(url, requestInit);
```
clebert: Please add a test, which checks that the requestInit is passed correctly to fetch.

niels-garve: Hi @clebert , I enhanced the existing one. Hope it works for you?

niels-garve:
Thanks @clebert and @unstubbable !

clebert:
Thank you for your contribution! A new version of the package @feature-hub/module-loader-commonjs (2.6.0) was published.

PR #586, opened 2020-05-25
fix(history-service): propagate history events if anchor is non empty

stemey:


packages/history-service/src/__tests__/index.test.ts
```diff
@@ -1387,6 +1387,36 @@ describe('defineHistoryService', () => {
...
 
+        it('calls listeners for consumer history changes if anchor is set', () => {
+          const listenerSpy = jest.fn();
+
+          historyService1.history.listen(listenerSpy);
+          historyService1.history.push('/bar#xxx');
+          historyService1.history.push('/bar#xxx');
+
+          expect(listenerSpy).toHaveBeenCalledTimes(2);
+
```
stemey, line 1398: This fails because listener is called 4 times instead of 2.

stemey:
There is an asymmetry between anchor handling and the rest of the url. There can only be one anchor and it should be changeable by any consumer. The event must go to the one implementing the anchor behavior. In our case it is currently the primary consumer but in general each feature app should be able to implement it by themselves. This is an open topic which should be discussed.

!!Anchor shouldn't be used for state management but only for scroll the page.

stemey:
> With this change, pushes/replaces of the same location won't lead to listener calls anymore. We need to talk about the implications.

It will only lead to listener calls if the location has changed or the anchor is set. Maybe that is not good enough. We might have to call listeners on every push/replace. The pop event could be sent to each consumer who has a none trivial location or to changed locations.

unstubbable:
Not needed anymore.

Unfortunately the listener implementation in the history service is not symmetrical. Therefore when pushing onto the primary history you also need to listen on the primary history to receive subsequent pushes/replaces with unchanged locations (including the hash). The same symmetry must be used when using the root history.

PR #589, opened 2020-06-12
feat(all): pass featureAppName as consumerName to Feature Service binder

unstubbable:


docs/guides/writing-a-feature-service.md
```diff
@@ -307,6 +308,36 @@ const myFeatureServiceDefinition = {
 };
 ```
 
+## When to use the `consumerName` instead of the `consumerId`
+
+In contrast to the `consumerId`, the `consumerName` must not be unique. It can
+be used for display purposes, logging, looking up Feature App configs, etc.,
+e.g. a Logger Feature Service could use it like this:
```
stemey, line 315: I would add here that a consumerName is usually unique for feature app's source url and can be used to look up meta data.

stemey: Maybe we should name consumerName featuerAppName insteadand it should be undefined if the consumer is a feature service. Seems better to me. It might actually be necessary to pass the featureAppName from one feature service to the next.

stemey:
We should continue to work on this.

stemey (approved):

PR #576, opened 2020-03-31
build(all): update TypeScript, TypeDoc, and Styled Components

unstubbable:
Check https://deploy-preview-576--feature-hub.netlify.com/@feature-hub/

closes #601

PR #612, opened 2020-10-31
GitHub Actions

unstubbable:
This replaces the Azure and Travis integrations for CI and releases with GitHub Actions.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,107 @@
...
+      CI: true
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 10.x
```
clebert, line 50: Why node 10?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Install
+        run: yarn install
+
+      - name: Lint
```
clebert, line 103: Is it really faster to do lint/compile/test as individual steps? instead of installing only once and then using npm-run-all -p ?

unstubbable: Yes. Remember that the installs are also done in parallel. This also gives us a cleaner output and failure notification compared to `run-p`.

.github/workflows/ci.yml
unstubbable: Previously we had node 8 and 10. But 8 is not supported by our current set of dependencies anymore. Because of this I already had to disable the engine check in 2c46feb of #576. With #611 we also won't be able to use 8 anymore because prettier needs at least 10.

We could think about adding newer versions using a matrix.

PS: Removing node 8 from the pipelines might technically be considered breaking, since we can't officially support it anymore. Should we officially make this a 3.0.0 release because of this? I'm hesitant to do this.

clebert: Sorry, I have not made myself clear.I am more interested in Node 12, which is supported by Lambda and is the default for aws-simple. Therefore we should test Node 12 and Node 10 as a matrix test.

.github/workflows/ci.yml
clebert: Ok

.github/workflows/ci.yml
unstubbable: Got you. 👌 Note: I had to update jest, babel and tslint to be fully compatible with node 12.

clebert (approved):

PR #613, opened 2020-11-12
fix(react): do not set state after FeatureAppContainer has been unmounted

unstubbable:
TODO:
- [x] add a unit test

### DIFF
--- a/packages/react/src/internal/internal-feature-app-container.tsx
+++ b/packages/react/src/internal/internal-feature-app-container.tsx
@@ -275,7 +275,9 @@ export class InternalFeatureAppContainer<
 
       featureApp.loadingPromise
         .then(() => {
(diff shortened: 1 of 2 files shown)


packages/react/src/internal/internal-feature-app-container.tsx
```diff
@@ -275,7 +275,9 @@ export class InternalFeatureAppContainer<
 
       featureApp.loadingPromise
         .then(() => {
-          this.setState({loading: false});
+          if (this.mounted) {
```
clebert, line 278: Maybe add a unit test for this

clebert (changes requested):

stemey (approved):

clebert (approved):

PR #629, opened 2021-02-02
Webpack Module federation

stemey:


packages/core/src/externals-validator.ts
```diff
@@ -45,6 +45,7 @@ export class ExternalsValidator {
    * @throws Throws an error if the required externals can't be satisfied.
    */
   public validate(requiredExternals: RequiredExternals): void {
+    if (true) return;
```
stemey, line 48: externals validator turned off temporarily. Can externals and shard modules gracefully be merged?

packages/core/src/feature-app-manager.ts
```diff
@@ -60,7 +60,7 @@ export interface FeatureAppDefinition<
   create(env: FeatureAppEnvironment<TFeatureServices, TConfig>): TFeatureApp;
 }
 
-export type ModuleLoader = (url: string) => Promise<unknown>;
+export type ModuleLoader = (url: string, scope?: string) => Promise<unknown>;
```
stemey, line 63: Scope must be a unique identifier for a feature app. It is the global name of the module. Currently the name id taken from the url of the module. Providing a meaningful name requires an API change. The id must b known to the feature app at build time and the integrator at runtime through feature app configuration.

packages/demos/src/module-loader-federation/webpack-config.js
```diff
@@ -0,0 +1,98 @@
...
+          use: 'url-loader',
+        },
+      ],
+    },
+    plugins: [
+      new ModuleFederationPlugin({
+        filename: 'remoteEntry.js',
+        name: 'remoteEntry',
+        exposes: {
+          './featureAppDefinition':
```
stemey, line 45: each expos will result in a new chunk whose name is derived from the path.

packages/demos/src/module-loader-federation/webpack-config.js
```diff
@@ -0,0 +1,98 @@
...
+        },
+        {
+          test: /\.(png|woff|woff2|eot|ttf|svg)$/,
+          use: 'url-loader',
+        },
+      ],
+    },
+    plugins: [
+      new ModuleFederationPlugin({
+        filename: 'remoteEntry.js',
```
stemey, line 42: the remote Entry contains the map of exposed modules and loads them on demand.

packages/module-loader-federation/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    const factory = await (window as any)[scope].get(module);
+    const Module = factory();
+    return Module;
+  };
+}
+
+export type WebpackInitSharing = () => any;
+export type WebpackShareScopes = {[scope: string]: {get: () => any}};
+
+export const loadFederationModuleFactory = (
```
stemey, line 25: webpack globals are passed into library code as they are rewritten by webpack.

packages/core/src/feature-app-manager.ts
stemey: scope can be set to a unique name, as long as the onload listener of the script is picking it up and removing it from the window. This is the way SystemJs works.

packages/demos/src/module-loader-federation/webpack-config.js
stemey: now using a unique global `featurehubGlobal` and exposing the standard `./featureAppDefinition`

packages/demos/src/node-integrator.ts
```diff
@@ -1,6 +1,6 @@
 import {Css} from '@feature-hub/react';
 import express from 'express';
-import webpack from 'webpack';
+// import webpack from 'webpack';
```
stemey, line 3: Typings not working with webpack 5 any more

unstubbable: we need https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50802 so that Webpack 5 types are compatible with the webpack-dev-middleware types

package.json
```diff
@@ -68,6 +68,7 @@
     "size-limit": "^4.9.2",
     "sort-package-json": "^1.16.0",
     "ts-config": "^20.9.0",
+    "babel-loader": "8.2.2",
```
stemey, line 71: babel loader needed to run int test "module-loader-federation" from root.

packages/demos/src/advanced-routing/webpack-config.js
```diff
@@ -1,4 +1,3 @@
-// @ts-check
```
stemey, line 1: removed because of webpack 5 issues

PR #631, opened 2021-02-05
fix(demos): watch node integrator

### LINKED ISSUE #250: Demos: Watcher does not work with node integrator

unstubbable:
By also bundling the node integrator with `webpack-dev-middleware` we can avoid having to restart the demo server when changes to server-side rendered demos are made.

This is based on https://github.com/webpack/webpack-dev-middleware#server-side-rendering.

fixes #250

### DIFF
--- a/packages/demos/src/custom-logging/index.test.ts
+++ b/packages/demos/src/custom-logging/index.test.ts
@@ -8,7 +8,7 @@ import {ConsoleMessage} from 'puppeteer';
 import {Browser} from '../browser';
 import {startServer} from '../start-server';
 import webpackConfigs from './webpack-config';
-import nodeWebpackConfig from './webpack-config.node';
+import nodeIntegratorWebpackConfig from './webpack-config.node';
 
 jest.setTimeout(120000);
 
@@ -19,7 +19,7 @@ describe('integration test: "custom logging"', () => {
   let consoleMessages: ConsoleMessage[];
 
   beforeAll(async () => {
-    server = await startServer(webpackConfigs, nodeWebpackConfig);
+    server = await startServer(webpackConfigs, nodeIntegratorWebpackConfig);
     consoleMessages = [];
 
     page.on('console', consoleMessages.push.bind(consoleMessages));
--- a/packages/demos/src/custom-logging/index.test.ts
+++ b/packages/demos/src/custom-logging/index.test.ts
@@ -18,7 +18,7 @@ describe('integration test: "custom logging"', () => {
   let consoleMessages: ConsoleMessage[];
 
   beforeAll(async () => {
-    server = await startServer(webpackConfigs, undefined);
+    server = await startServer(webpackConfigs);
     consoleMessages = [];
 
     page.on('console', consoleMessages.push.bind(consoleMessages));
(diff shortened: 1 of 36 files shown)


packages/demos/src/start-server.ts
```diff
@@ -70,14 +59,22 @@ function createDocumentHtml(
...
   webpackConfigs: webpack.Configuration[],
-  renderApp: AppRenderer | undefined,
   demoName?: string
 ): Promise<Server> {
   const port = await getPort(demoName ? {port: 3000} : undefined);
   const app = express();
 
+  for (const compiler of webpack(webpackConfigs).compilers) {
+    const serverSideRender =
+      compiler.options.output?.filename === nodeIntegratorFilename;
```
stemey, line 69: from my point of view multiple dedicated webpack configs are better. Either two arrays for client and server or an extra wrapper with boolean property. This checking of the fileName is an implicit dependency, that is brittle. If more than two webpacks are needed

unstubbable: I'm not sure I completely understand what you are suggesting here, since there can only be one node integrator per demo. But you do have a point that hard coding the node integrator file name might not be sich a good idea. I tried to improve on this in #639. Please have a look.

packages/demos/src/custom-logging/index.test.ts
```diff
@@ -18,7 +19,7 @@ describe('integration test: "custom logging"', () => {
   let consoleMessages: ConsoleMessage[];
 
   beforeAll(async () => {
-    server = await startServer(webpackConfigs);
+    server = await startServer(webpackConfigs, nodeWebpackConfig);
```
clebert, line 22: Maybe you could also use an implicit detection via `target: 'node'` instead of this in my opinion unattractive API? This way the user would not have to be burdened with it.

clebert, line 22: Maybe you could also use an implicit detection via `target: 'node'` instead of this in my opinion unattractive API? This way the user would not have to be burdened with it.

unstubbable: Hm, the `node` target is also used for the feature app commonjs bundles. I can’t think of a unique discrimitor.

packages/demos/src/node-integrator.ts
```diff
@@ -0,0 +1,50 @@
...
+import express from 'express';
+import webpack from 'webpack';
+
+export interface AppRendererOptions {
+  port: number;
+  req: express.Request;
+}
+
+export interface AppRendererResult {
+  html: string;
```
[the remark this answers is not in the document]
unstubbable: yes, will add it (This is old code I just moved here.)

packages/demos/src/custom-logging/index.test.ts
unstubbable: Also, there is either none or one node integegrator. Therefore a differently structured item in the list of webpack configs is not the best design, I think. It should be its own entity.

unstubbable: By the way, maybe the special meaning would have been clearer if I named it `nodeIntegratorWebpackConfig` instead of `nodeWebpackConfig`?

clebert (approved):

pachecocordovamoiseseduardo-byte (approved):

PR #647, opened 2021-02-27
feat(demos): update to webpack 5

unstubbable:


packages/demos/package.json
```diff
@@ -3,6 +3,13 @@
   "version": "2.8.1",
   "private": true,
   "license": "MIT",
+  "workspaces": {
+    "nohoist": [
+      "@types/webpack-dev-middleware",
+      "webpack",
+      "ts-loader"
+    ]
+  },
```
unstubbable, line 12: Our augmented types for `webpack` (`packages/demos/types/webpack.d.ts`) are not respected by TypeScript, unless we prevent the two packages that need them (as well as webpack itself) from being hoisted to the root.

.nvmrc
```diff
@@ -1 +1 @@
-10
+12
```
clebert, line 1: I think we can remove Node.js support for v10 (2018) from CI as well. Maintenance end of that version is in a little more than a month: 2021-04-30

packages/demos/src/webpack-base-config.js
```diff
@@ -43,6 +43,7 @@ const webpackBaseConfig = {
   },
   plugins: [
     new webpack.DefinePlugin({
+      'process.env.BLUEPRINT_NAMESPACE': '""',
```
clebert, line 46: Why is the ENV var new?

.nvmrc
unstubbable: Let's do that in another PR.

packages/demos/src/webpack-base-config.js
unstubbable: [Webpack 5 does no longer include a polyfill for `process`](https://webpack.js.org/migrate/5/#run-a-single-build-and-follow-advises). This env variable was/is accessed by blueprint (and they fall back to a default value if it is falsy).

.nvmrc
clebert: See, #648

clebert (approved):

PR #685, opened 2021-07-01
Handle module type during hydration

unstubbable:
- fix(all): handle module type during hydration
- refactor(core): extract error message

### DIFF
--- a/packages/demos/src/server-side-rendering/webpack-config.js
+++ b/packages/demos/src/server-side-rendering/webpack-config.js
@@ -6,33 +6,49 @@ const {merge} = require('webpack-merge');
 const {webpackBaseConfig} = require('../webpack-base-config');
 
 /**
- * @type {webpack.Configuration}
- */
-const featureAppConfig = merge(webpackBaseConfig, {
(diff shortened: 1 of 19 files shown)


packages/demos/src/server-side-rendering/webpack-config.js
```diff
@@ -5,34 +5,50 @@ const webpack = require('webpack');
...
+    },
+    plugins: [
+      new webpack.container.ModuleFederationPlugin({
+        filename: 'feature-app.federated.js',
+        name: '__feature_hub_feature_app_module_container__',
+        exposes: {
+          featureAppModule: path.join(__dirname, './feature-app'),
+        },
+        shared: {
+          react: {singleton: true},
```
clebert, line 36: why is wmf not also configured for the integrator?

clebert (changes requested):

packages/demos/src/server-side-rendering/webpack-config.js
unstubbable: fixed

clebert (approved):

PR #687, opened 2021-07-06
build(all): switch website hosting from Netlify to AWS

unstubbable:
I have already changed the DNS setup and tested the deployment locally. So https://feature-hub.io/ is now hosted on AWS. This PR adds the necessary `aws-simple` config and GitHub Actions workflows to set up the deployment for the CI (including PR deployments). I also had to change the API docs path from `@feature-hub` to `api` because AWS does not allow an `@` in an API Gateway route.

Note: The command actions can not be tested in this PR, because they must be available in `master` first.


unstubbable:
The Netlify jobs are expected to fail. We can ignore those. I've now deactivated the builds on netlify.com, so I think they won't appear in future PRs. If they do, we need to remove the Netlify GitHub integration for this repo at the org level.

PR #693, opened 2021-07-09
Document Bundling

unstubbable:
- https://bbc112bb.feature-hub.io/docs/guides/writing-a-feature-app
- https://bbc112bb.feature-hub.io/docs/guides/sharing-npm-dependencies

### DIFF
--- a/docs/guides/sharing-npm-dependencies.md
+++ b/docs/guides/sharing-npm-dependencies.md
@@ -1,11 +1,26 @@
 ---
 id: sharing-npm-dependencies
-title: Sharing NPM Dependencies
-sidebar_label: Sharing NPM Dependencies
+title: Sharing npm Dependencies
+sidebar_label: Sharing npm Dependencies
 ---
 
-> The feature described in this guide is also demonstrated in the ["AMD Module
+The integrator might want to share commonly used libraries with Feature Apps for
+two reasons.
+
+1. Reduce the bundle size of the Feature Apps by omitting shared dependencies
+   from their bundle.
+1. Ensure that a library that can only exist once on a page is provided as a
+   singleton.
+
+How dependencies are shared depends on the [provided module loaders of the
+integrator][module-loader].
+
+## AMD
+
+> The feature described in this section is also demonstrated in the ["AMD Module
 > Loader" demo][amd-module-loader-demo].
+
+### Integrator
 
 When using the [`@feature-hub/module-loader-amd`][module-loader-amd-api]
 package, the integrator can provide shared npm dependencies to Feature Apps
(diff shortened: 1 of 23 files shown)


docs/guides/sharing-npm-dependencies.md
```diff
@@ -25,19 +40,122 @@ const {featureAppManager} = createFeatureHub('acme:integrator', {
...
+If a Feature App wants to provide a shared module, it should _not_ be `eager`.
+In case of a version mismatch (defined by the [semver][] ranges in the
+`package.json` files of the Feature Apps) the module is loaded multiple times,
+and available to other Feature Apps in each loaded version.
+
+> **Note**:  
+> A large library of which only a small portion is used, might not be a good
+> candidate to provide as a shared module. In this case it might be more
+> efficient to include the library in each Feature App bundle, while making use
+> of [Webpack's tree shaking capabilities][tree-shaking].
```
stemey, line 148: Maybe we could add here: It is more efficient to share a custom wrapper of a library providing a common set of use cases for the feature apps. Thus, the bundle size can be reduced by tree shaking.

stemey: we could also talk about the approach taken by OneShop to load the library lazily. This is done for feature services that are only used in the client. It should be possible for client side rendered feature apps also.

docs/guides/writing-a-feature-app.md
```diff
@@ -122,20 +122,31 @@ properties:
...
+   configuration meta data, etc.
 
 1. `baseUrl` — A base URL to be used for referencing the Feature App's own
-   resources. It is only set in the `env` if the integrator has defined a
+   resources. It is only set in the `env` if the integrator^1^ has defined a
    `baseUrl` on the corresponding
    [`FeatureAppLoader`][feature-app-loader-base-url] or
    [`FeatureAppContainer`][feature-app-container-base-url].
 
+1. `done` — A callback that the integrator^1^ might have defined. A short-lived
```
stemey, line 142: we need an example for done. This is very use case specific and not as generic as the other properties.

docs/guides/sharing-npm-dependencies.md
unstubbable: Should we maybe add these details to the "Reducing the Bundle Size" guide and cross reference it from here?

docs/guides/writing-a-feature-app.md
unstubbable: How about this?

> A short-lived Feature App can call this function when it has completed its task, thus giving the integrator a hint, that it can be removed. For example, if the Feature App was opened in a layer, the integrator could close the layer when `done()` is called.

unstubbable:
@stemey Please have another look at these docs:

- https://167d9ea9.feature-hub.io/docs/guides/integrating-the-feature-hub#done
- https://167d9ea9.feature-hub.io/docs/guides/integrating-the-feature-hub#done-1
- https://167d9ea9.feature-hub.io/docs/guides/writing-a-feature-app#create (`done`)
- https://167d9ea9.feature-hub.io/docs/guides/sharing-npm-dependencies#feature-apps-1
- https://167d9ea9.feature-hub.io/docs/guides/reducing-the-bundle-size

stemey (approved):

PR #701, opened 2022-01-10
feat(all): add failure option to done call

elbkind:
With this a feature app can include some other FA to select
something and the new FA can signal in the done call if
changes should be applied or if it just terminated and
nothing should be changed

### DIFF
--- a/packages/core/src/feature-app-manager.ts
+++ b/packages/core/src/feature-app-manager.ts
@@ -45,8 +45,11 @@ export interface FeatureAppEnvironment<
    * If this callback is defined, a short-lived Feature App can call this
    * function when it has completed its task. The Integrator (or parent Feature
    * App) can then decide to e.g. unmount the Feature App.
+   *
+   * By default it is assumed that the FA exited successfully, but it can signal
+   * that it failed. It is up to the integrator to handle this.
    */
-  readonly done?: () => void;
+  readonly done?: (failed?: boolean) => void;
 }
 
 export interface FeatureAppDefinition<
@@ -118,8 +121,11 @@ export interface FeatureAppScopeOptions<
    * short-lived Feature App can call this function when it has completed its
    * task. The Integrator (or parent Feature App) can then decide to e.g.
    * unmount the Feature App.
+   *
+   * By default it is assumed that the FA exited successfully, but it can signal
+   * that it failed. It is up to the integrator to handle this.
    */
-  readonly done?: () => void;
+  readonly done?: (failed?: boolean) => void;
 }
 
 export interface FeatureAppManagerOptions {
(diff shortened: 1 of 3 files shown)


packages/core/src/feature-app-manager.ts
```diff
@@ -45,8 +45,11 @@ export interface FeatureAppEnvironment<
    * If this callback is defined, a short-lived Feature App can call this
    * function when it has completed its task. The Integrator (or parent Feature
    * App) can then decide to e.g. unmount the Feature App.
+   *
+   * By default it is assumed that the FA exited successfully, but it can signal
+   * that it failed. It is up to the integrator to handle this.
    */
-  readonly done?: () => void;
+  readonly done?: (failed?: boolean) => void;
```
unstubbable, line 52: For scalability/flexibility reasons I would prefer to use an unknown result parameter instead, i.e.:

```suggestion
  readonly done?: (result?: unknown) => void;
```

The type of the `result` can then either be casted, or type predicates might be used to [narrow](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) the type at runtime.

unstubbable (changes requested):

packages/core/src/feature-app-manager.ts
elbkind: I am not sure, initially i was thinking about a more elaborate parameter, but this would tie integrator and FA more tightly than i care about.
So maybe `"ok" | unknown` with unknown always being an error indicator?

packages/core/src/feature-app-manager.ts
```diff
@@ -45,8 +45,11 @@ export interface FeatureAppEnvironment<
    * If this callback is defined, a short-lived Feature App can call this
    * function when it has completed its task. The Integrator (or parent Feature
    * App) can then decide to e.g. unmount the Feature App.
+   *
+   * By default it is assumed that the FA exited successfully, but it can signal
+   * that it failed. It is up to the integrator to handle this.
```
unstubbable, line 50: We should update this text now. Suggestion:

> Optionally, the Feature App can pass a `result` into the `done` callback. The type/structure of the `result` must be agreed between the Integrator (or parent Feature App) and the Feature App.

unstubbable (changes requested):

packages/core/src/feature-app-manager.ts
elbkind: done

unstubbable: Thanks. Can you please formate the comments to use 80 characters? I can recommend https://marketplace.visualstudio.com/items?itemName=stkb.rewrap for this task. And please also rebase the branch.

unstubbable (approved):

PR #704, opened 2022-02-10
fix(history-service): compatibility with react-router-dom

unstubbable:
The methods of `history.History` must be bound explicitly, because components like `Link` from the `react-router-dom` package deconstruct methods like `push` and `replace` from the `history`.


janro1 (approved):
LGTM

packages/history-service/src/internal/consumer-history.ts
```diff
@@ -17,6 +17,19 @@ export abstract class ConsumerHistory implements history.History {
...
+    /**
+     * The methods of `history.History` must be bound explicitly, because
+     * components like `Link` from the `react-router-dom` package deconstruct
+     * methods like `push` and `replace` from the `history`.
+     */
+    this.push = this.push.bind(this);
+    this.replace = this.replace.bind(this);
+    this.go = this.go.bind(this);
+    this.goBack = this.goBack.bind(this);
+    this.goForward = this.goForward.bind(this);
```
nikmilson, line 30: BTW, could you explain, why do you define `goBack` and `goForward`, while [history](https://github.com/remix-run/history) defines [back](https://github.com/remix-run/history/blob/ba1cd3cf50465af8e1a89cc4f088c6372c10c12b/docs/api-reference.md#historyback) and [forward](https://github.com/remix-run/history/blob/ba1cd3cf50465af8e1a89cc4f088c6372c10c12b/docs/api-reference.md#historyforward)?

clebert (approved):

PR #710, opened 2022-03-31
feat(core): introduce onBind callback

unstubbable:
supersedes and closes #708

### DIFF
--- a/packages/core/src/feature-app-manager.ts
+++ b/packages/core/src/feature-app-manager.ts
@@ -130,6 +130,12 @@ export interface FeatureAppScopeOptions<
   readonly done?: (result?: unknown) => void;
 }
 
+export interface OnBindParams {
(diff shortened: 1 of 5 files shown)


clebert (approved):
+1

packages/core/src/feature-app-manager.ts
```diff
@@ -432,10 +455,12 @@ export class FeatureAppManager {
     };
 
     try {
-      if (beforeCreate) {
-        beforeCreate(env);
-      }
+      beforeCreate?.(env);
```
stemey, line 458: this changes the current behavior. Is this intended?

unstubbable: It was, but after discussing it I agree that this should be reverted.

stemey (approved):

PR #714, opened 2022-07-06
feat(core): support external validator to be supplied in options

stemey:


unstubbable:
This PR has a couple of minor issues and also an accidental breaking change in it that we can talk about in detail. But I'm wondering whether we even need to support this kind of customisation in `createFeatureHub`. After all, this is only a convenience function that is not necessary to use. Specifically for this use case you might replace a `creatFeatureHub` call with only a couple of statements that already allow you to customise the externals validator.

E.g. before:

```ts
const featureHub = createFeatureHub(integratorId, {
  moduleLoader,
  featureServiceDefinitions,
  featureServiceDependencies,
  providedExternals,
  logger,
});
```

after:

```ts
const externalsValidator = new CustomExternalsValidator(providedExternals);
const featureServiceRegistry = new FeatureServiceRegistry({externalsValidator, logger});

featureServiceRegistry.registerFeatureServices(featureServiceDefinitions, integratorId);

const featureAppManager = new FeatureAppManager(featureServiceRegistry, {
  externalsValidator,
  moduleLoader,
  logger,
});

const {featureServices} = featureServiceRegistry.bindFeatureServices(
  {dependencies: {featureServices: featureServiceDependencies}},
  integratorId
);

const featureHub: FeatureHub = {featureAppManager, featureServiceRegistry, featureServices};
```

assuming you've created a custom externals validator like this:

```ts
class CustomExternalsValidator extends ExternalsValidator {
  constructor(providedExternals: ProvidedExternals) {
    super(providedExternals);
  }

  validate(requiredExternals: RequiredExternals): void {
    // custom validation
  }
}
```

Furthermore, the externals validator is also a feature that we don't want to support longterm in the Feature Hub, especially because the webpack module federation loader is now available.

PR #716, opened 2022-09-06
feat(history-service): introduce v3 of the history service

unstubbable:
Version 3 of the history service provides a `history` that is compatible with v5 of the [history package](https://www.npmjs.com/package/history) and therefore can also be used together with v6 of [React Router](https://reactrouter.com).

BREAKING CHANGE: Version 5 of the `history` package is now required as peer dependency. Some types of the root location transformer have been slightly changed, and the deprecated `primaryConsumerId` has been removed (replaced by `primaryConsumerHistoryKey`).


clebert (dismissed):

docs/guides/sharing-the-browser-history.md
```diff
@@ -61,18 +61,18 @@ const myFeatureAppDefinition = {
...
           <App />
-        </Router>
+        </HistoryRouter>
       ),
     };
   },
 };
 ```
 
 The `history` property of the History Service is API-compatible with the history
```
stemey, line 73: I think exposing a foreign API (history) that cannot be implemented completely and changes with the external dependency changes, wasn't a good idea. We should replace the API with a feature-hub API. LocationObject/ConsumerLocation was always a hassle and now has changed. History has changed although not completely supported.

unstubbable: How do you want to support the [react-router use case](https://github.com/sinnerschrader/feature-hub/blob/864611d40ff51b6ece4c0e3d40827141bf39b001/docs/guides/sharing-the-browser-history.md#as-a-feature-app) then? Can you also elaborate what exactly the hassle was so that we can potentially improve on that?

clebert (approved):

PR #726, opened 2023-02-21
feat(all): support multiple versions od feature service

stemey:


crebelskydico:
@stemey Is there any possibility that this is released soon? Currently i'm facing this issue in a project and it would help a lot. Tell me how i could help if help is needed. Thanks in advance.

PR #742, opened 2023-09-21
feat(all): pass `parentFeatureApp` into `onBind`

unstubbable:
This allows the integrator to collect information about "Feature App in Feature App" usage.

### DIFF
--- a/packages/react/src/internal/internal-feature-app-container.tsx
+++ b/packages/react/src/internal/internal-feature-app-container.tsx
@@ -10,6 +10,7 @@ import {
   FeatureApp,
 } from '../feature-app-container';
 import {FeatureHubContextConsumerValue} from '../feature-hub-context';
+import {FeatureAppContext} from './feature-app-context';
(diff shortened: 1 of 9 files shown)


stemey (approved):

packages/react/src/internal/internal-feature-app-container.tsx
```diff
@@ -24,11 +25,7 @@ export const handleError = (
   }
 };
 
-export interface BaseFeatureAppContainerProps<
-  TFeatureApp,
-  TFeatureServices extends FeatureServices = FeatureServices,
-  TConfig = unknown
-> {
+export interface FeatureAppDescriptor {
```
clebert, line 28: Maybe import the interface from `packages/core/src/feature-app-manager.ts`

unstubbable: I didn't do that because it has (intentionally) different jsdoc comments for the properties.

clebert (approved):

PR #762, opened 2023-09-28
build(all): speed up demos and itests by using `swc-loader`

unstubbable:
Normally [SWC does not support project references](https://github.com/swc-project/swc/discussions/2156), but by specifying the source files in `exports`, using a special condition name, we can trick webpack into resolving the source files, instead of the JS files from `lib`, and therefore letting SWC transpile the sources. With this, we can move away from `ts-loader`, and still avoid compilation with `tsc` as prebuilt step.

Inspired by https://github.com/evanw/esbuild/issues/1250#issuecomment-1463826174.


clebert (commented):
Excellent solution. I wonder, is "internal-bundler" a commonly used nomenclature for this condition? Does an "external bundler" exist too? Perhaps it could be named "swc" or just "bundler".

clebert:
Ok understood. Makes sense.

clebert (approved):

PR #770, opened 2023-12-21
feat(async-ssr-manager): support concurrent `renderUntilCompleted` calls

unstubbable:
When the integrator is not a React app, and renders multiple feature apps concurrently with separate `ReactDOM.renderToString` calls, each wrapped with `renderUntilCompleted`, we must scope the scheduled async operations to the respective `renderUntilCompleted` call. This is accomplished by leveraging `AsyncLocalStorage`.

Note: This scenario did kind of work already, by accident, up until v2.9.0. But it was broken with #702 which we shipped with [v2.10.0](https://github.com/sinnerschrader/feature-hub/releases/tag/v2.10.0).


packages/async-ssr-manager/src/__tests__/index.test.ts
```diff
@@ -249,6 +249,69 @@ describe('defineAsyncSsrManager', () => {
...
+
+          expect(html).toEqual([
+            'render call 0, pass 2',
+            'render call 1, pass 2',
+          ]);
+        });
+      });
+
+      describe('with a consumer that schedules a rerender outside of `renderUntilCompleted`', () => {
+        it('resolves with an html string after the second render pass', async () => {
```
clebert, line 297: This description seems to be wrong

clebert (changes requested):

packages/async-ssr-manager/src/__tests__/index.test.ts
unstubbable: true, will fix

clebert (approved):

PR #785, opened 2024-11-15
feat(react): handle cleanup of DOM feature app on it's removal

msaktor:
Hi! First thanks a lot for this amazing library. It allowed us to greatly scale frontend development at Jamf.

We would like to suggest a possibility to do a cleanup when `feature-app-container` is disconnected from the DOM, in order to avoid memory leaks.

Here I present a suggestion that uses a return function in `attachTo` in a similar fashion how "useEffect" works in React. The benefit is access to variables declared inside the `attachTo` scope.

Another solution could be a `detachFrom` function that can be an optional fn on `DomFeatureApp` interface.

Example for React:
```tsx
const featureAppDefinition: FeatureAppDefinition<DomFeatureApp> = {
  dependencies: {
    featureServices: {},
  },
  create: (env: FeatureAppEnvironment<FeatureServices, Config>) => {
    return {
      attachTo(element: HTMLElement): void {
        reactRoot = createRoot(element);
        reactRoot.render(<div>Hello world!</div>);

        return () => {
          reactRoot.unmount();
        };
      },
    };
  },
};
```

Or Vue:
```ts
const featureAppDefinition: FeatureAppDefinition<DomFeatureApp> = {
  dependencies: {
    featureServices: {},
  },
  create: (env: FeatureAppEnvironment<FeatureServices, Config>) => {
    return {
      attachTo(element: HTMLElement): void {
        const app = Vue.createApp({
          template: '<div>Hello world!</div>',
        });
        app.mount('#app');

        return () => {
           app.unmount();
        };
      },
    };
  },
};
```

### DIFF
--- a/docs/guides/writing-a-feature-app.md
+++ b/docs/guides/writing-a-feature-app.md
@@ -235,6 +235,34 @@ const myFeatureAppDefinition = {
 };
 ```
 
+The `attachTo` method of a DOM Feature App can optionally return a "detach"
+function. This function is invoked when the Feature App is removed from the DOM,
+allowing it to perform necessary cleanup tasks (e.g. unsubscribing from events).
+This behavior is similar to how the `useEffect` hook works in React.
+
+```js
+const myFeatureAppDefinition = {
+  create(env) {
+    return {
+      attachTo(container) {
+        const app = Vue.createApp({
+          template: '<div>Hello world!</div>',
+        });
+        app.mount(container);
+
+        return function () {
+          // unsubscribe from events
+          window.removeEventListener('resize', handleResize);
+
+          // or unmount apps
+          app.unmount();
+        };
+      },
+    };
+  },
+};
+```
+
 ### Loading UIs provided by the React Integrator
 
 Both kinds of Feature Apps can specify a loading stage for Feature Apps, which
--- a/packages/demos/src/integrator-dom/feature-app.ts
+++ b/packages/demos/src/integrator-dom/feature-app.ts
@@ -6,7 +6,7 @@ const featureAppDefinition: FeatureAppDefinition<DomFeatureApp> = {
     attachTo(element: HTMLElement): DetachFunction {
       element.replaceWith('Hello, World!');
 
-      return function () {
+      return () => {
         document.title = 'Detached!';
       };
     },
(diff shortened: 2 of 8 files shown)


NSteinhoff:
Hi @msaktor, thank you for your contribution! 

I think this is a good improvement. I also agree with your reasoning to prefer a cleanup callback à la `useEffect` over a `detachFrom` method.

---

One issue I see, is that a DOM feature app can also be rendered by a React `FeatureAppContainer`, which would currently not run the cleanup.

https://github.com/feature-hub/feature-hub/blob/11490453001ea2628ecb9aa291d494618f429305/packages/react/src/feature-app-container.tsx#L39

https://github.com/feature-hub/feature-hub/blob/11490453001ea2628ecb9aa291d494618f429305/packages/react/src/internal/internal-feature-app-container.tsx#L326

I think we should ensure consistent behavior of DOM feature apps for both DOM and React containers.

---

We currently have limited capacity to invest into active feature-hub development. You could help this along greatly by providing the appropriate documentation/demo/test updates. We'll try to get to it as soon as we are able.

Thanks,
Niko

msaktor:
Hi Niko @NSteinhoff !

I've updated the PR with tests and examples, can you have another look please?

Thanks!
Marek

NSteinhoff (dismissed):
Hi @msaktor, thanks for the update! Looks great 👍 

We need to investigate and fix the failing builds. It's possible that they are not related to your changes, but I'm not yet familiar enough with the CI setup to be sure. Are you blocked by this?

NSteinhoff (approved):
Hi @msaktor, I took the liberty to update the PR to meet the linting requirements. It's now good to go and I will merge it shortly. Thanks again for your contribution! 🥇 

The coverage action fails for PRs from forks, because it does not have the necessary rights to post the report. Good to know and we'll probably want to address it, but we can ignore it for this PR.

msaktor:
Thanks a lot for merging this Niko! @NSteinhoff 

I see that npm still has the latest version 3.7.0 and not 3.8.0, are you planning to publish the new version soon?

NSteinhoff:
Hi @msaktor, the `@feature-hub/react` and `@feature-hub/dom` packages should be at 3.8.0:

https://www.npmjs.com/search?q=feature-hub

msaktor:
my bad, I was looking at @feature-hub/core 🤦 
thanks!

pachecocordovamoiseseduardo-byte (approved):

PR #787, opened 2024-11-28
fix(history-service): keep pathname if only search is defined in path

stemey:


NSteinhoff (commented):
👍 The code looks good, but I think the tests could be improved.

The tests are quite thorough, but somewhat hard to read, because the differences in the test cases are so far apart and hidden within the boilerplate. I think this screams for a table based test via `.each()`, which would collapse the full suite into 6 table rows. `push|replace x full|search|fragment`.

https://jestjs.io/docs/api#testeachtablename-fn-timeout

It's not quite clear to me why we couldn't start each test with the same initial location. That would enable all setup to be moved to a `beforeEach`, and would probably make the `createHistory` helper unnecessary. 

If you'd prefer to stick to the current test organization, then I think that they could use some white space to separate tests and top level blocks.

PR #789, opened 2025-08-15
feat(async-ssr-manager): add option ignoring render failures

kmendzie:
solves: https://github.com/feature-hub/feature-hub/issues/786

**Solution**
Resolve the `renderUntilCompleted()` of the `AsyncSsrManager` successfully even when a loadingPromise is rejected, in order to keep the ancestor react tree alive. This behaviour was made configurable via `AsyncSsrManagerOptions`, in order to avoid a breaking change. Furthermore the `addUrlForHydration` is only being called when the SSR loading actually succeeds since the hydration would be pointless without actual SSR markup.

**Notes:**
- avoids a breaking change
- css will not be touched to keep it minimally invasive

### DIFF
--- a/packages/async-ssr-manager/src/internal/async-ssr-manager.ts
+++ b/packages/async-ssr-manager/src/internal/async-ssr-manager.ts
@@ -74,7 +74,21 @@ export class AsyncSsrManager implements AsyncSsrManagerV1 {
         asyncOperations.clear();
 
         if (skipRenderError) {
-          await Promise.allSettled(asyncOperationsSnapshot);
+          await Promise.allSettled(asyncOperationsSnapshot).then(
+            (settledResults) => {
+              settledResults
+                .filter(
+                  (rejectedResults): rejectedResults is PromiseRejectedResult =>
+                    rejectedResults.status === 'rejected',
+                )
+                .forEach((rejected, index) => {
+                  this.context.logger.error(
+                    `Operation ${index} failed rendering with reason:`,
+                    rejected.reason,
+                  );
+                });
+            },
+          );
         } else {
           await Promise.all(asyncOperationsSnapshot);
         }
--- a/packages/async-ssr-manager/src/internal/async-ssr-manager.ts
+++ b/packages/async-ssr-manager/src/internal/async-ssr-manager.ts
@@ -1,5 +1,5 @@
 import {AsyncLocalStorage} from 'async_hooks';
-import {AsyncSsrManagerV1} from '..';
+import {AsyncSsrManagerOptions, AsyncSsrManagerV1} from '..';
 import {AsyncSsrManagerContext} from './async-ssr-manager-context';
 import {setTimeoutAsync} from './set-timeout-async';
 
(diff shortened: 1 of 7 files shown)


packages/async-ssr-manager/src/internal/async-ssr-manager.ts
```diff
@@ -16,7 +16,7 @@ export class AsyncSsrManager implements AsyncSsrManagerV1 {
 
   public constructor(
     private readonly context: AsyncSsrManagerContext,
-    private readonly timeout?: number,
+    private readonly options: AsyncSsrManagerOptions,
```
stemey: options should be optional

packages/react/src/__tests__/feature-app-loader.node.test.tsx
```diff
@@ -144,10 +144,7 @@ describe('FeatureAppLoader (on Node.js)', () => {
           />,
         );
 
-        expect(mockAddUrlForHydration).toHaveBeenCalledWith(
-          'http://example.com/example.js',
-          'a',
-        );
+        expect(mockAddUrlForHydration).not.toHaveBeenCalled();
```
stemey: Why is the assertion not inverted?

packages/react/src/__tests__/feature-app-loader.node.test.tsx
```diff
@@ -261,21 +255,41 @@ describe('FeatureAppLoader (on Node.js)', () => {
...
             <FeatureAppLoader
               featureAppId="testId"
               src="example.js"
               serverSrc="example-node.js"
+              css={[{href: 'foo.css'}]}
             />,
           );
         } catch {}
 
-        expect(mockAddUrlForHydration).toHaveBeenCalledWith(
```
stemey: Why is the assertion not inverted?

packages/react/src/feature-app-loader.tsx
```diff
@@ -169,6 +165,13 @@ class InternalFeatureAppLoader<TConfig = unknown> extends React.PureComponent<
       this.state = {error};
     } else if (featureAppDefinition) {
       this.state = {featureAppDefinition};
+
```
stemey, line 168: either this code needs to be added to the last condition as well or we need to improve the tests. We don't sem to have a test checking that url is added for hydration.

kmendzie: I have adjusted the test to allow using a synchronously set feature app definition.

packages/async-ssr-manager/src/internal/async-ssr-manager.ts
```diff
@@ -69,7 +73,11 @@ export class AsyncSsrManager implements AsyncSsrManagerV1 {
 
         asyncOperations.clear();
 
-        await Promise.all(asyncOperationsSnapshot);
+        if (skipRenderError) {
+          await Promise.allSettled(asyncOperationsSnapshot);
```
stemey: there needs to be logging about failures here.

kmendzie: Logging was added: https://github.com/feature-hub/feature-hub/pull/789/commits/174f90fe58b74b91494f072f5c1d1d7eb4ad01f4

packages/async-ssr-manager/src/internal/async-ssr-manager.ts
kmendzie: The options have been made optional: https://github.com/feature-hub/feature-hub/pull/789/commits/a010f0cf1c6ca7ec487d367385df0ebf535a515e

kmendzie:
Will be solved by https://github.com/feature-hub/feature-hub/pull/790

Issue #786, opened 2024-11-26
AsyncSsrManager + FeatureAppLoader: Module fetch error crashes rendering


soanaphikeste:
When loading the serverSrc fails during server side rendering, the `renderUntilCompleted` of the AsyncSsrManager rejects and rendering of the whole React tree fails.

**Steps to reproduce:**
1. Take any integrator capable of SSR (e.g. [the SSR demo](https://github.com/feature-hub/feature-hub/tree/main/packages/demos/src/server-side-rendering) in this repo) 
2. Specify a URL that returns invalid JS as the [serverSrc](https://github.com/feature-hub/feature-hub/blob/main/packages/demos/src/server-side-rendering/app.tsx#L15) (e.g. a URL that returns a 404 page)
3. Render page

**Actual result:**
<details><summary>Details</summary>

:heavy_check_mark: FeatureAppLoader [creates Feature App asynchronously](https://github.com/feature-hub/feature-hub/blob/main/packages/react/src/feature-app-loader.tsx#L164)
:heavy_check_mark: Because serverSrc is not loaded yet, FeatureAppLoader [registers loading promise in AsyncSsrManager](https://github.com/feature-hub/feature-hub/blob/main/packages/react/src/feature-app-loader.tsx#L173)
:heavy_check_mark: serverSrc is loaded and module loader throws an error [because of invalid JS](https://github.com/feature-hub/feature-hub/blob/main/packages/module-loader-commonjs/src/index.ts#L33) in serverSrc
:heavy_check_mark: Feature App Registry [rejects async creation promise](https://github.com/feature-hub/feature-hub/blob/main/packages/core/src/feature-app-manager.ts#L347)
:x: [AsyncSsrManager rendering loop fails](https://github.com/feature-hub/feature-hub/blob/main/packages/async-ssr-manager/src/internal/async-ssr-manager.ts#L72)

</details> 

:x: None of the Feature Apps on the page are rendered on server side. Instead an error is displayed.

**Expected result:**
:heavy_check_mark:  Failing to load the serverSrc should be treated as any other error that occurs while rendering the Feature App. The error should only affect the failed Feature App. 
:heavy_check_mark: All other Feature Apps on the page should continue with server side rendering as normal.

**Note:**
[...]

PR #790, opened 2025-09-29
feat(async-ssr-manager): split content and source rendering queues

kmendzie:
solves: https://github.com/feature-hub/feature-hub/issues/786

**Solution**
Enables an alternative internal method to schedule asyncOperations, that will be resolved in a separate rendering queue (`jsLoading`). Promises scheduled here may also include rejected promises to allow handling their error cases gracefully.

Furthermore the `addUrlForHydration` is used when the SSR loading actually succeeds only.

### DIFF
--- a/packages/async-ssr-manager/src/internal/async-ssr-manager.ts
+++ b/packages/async-ssr-manager/src/internal/async-ssr-manager.ts
@@ -88,43 +88,35 @@ export class AsyncSsrManager implements AsyncSsrManagerV1 {
     while (inProgress) {
       inProgress = false;
 
-      while (asyncOperations.contentLoading.size > 0) {
-        while (asyncOperations.contentLoading.size > 0) {
-          inProgress = true;
+      for (const queue of Object.keys(
+        asyncOperations,
+      ) as (keyof AsyncOperationQueues)[]) {
+        while (asyncOperations[queue].size > 0) {
+          while (asyncOperations[queue].size > 0) {
+            inProgress = true;
 
-          // Storing a snapshot of the asynchronous operations and clearing them
-          // afterwards, allows that consecutive promises can be added while the
-          // current asynchronous operations are running.
+            // Storing a snapshot of the asynchronous operations and clearing them
+            // afterwards, allows that consecutive promises can be added while the
+            // current asynchronous operations are running.
 
-          const asyncOperationsSnapshot = Array.from(
-            asyncOperations.contentLoading.values(),
-          );
+            const asyncOperationsSnapshot = Array.from(
+              asyncOperations[queue].values(),
+            );
 
-          asyncOperations.contentLoading.clear();
+            asyncOperations[queue].clear();
 
-          await Promise.all(asyncOperationsSnapshot);
+            if (queue === 'contentLoading') {
+              await Promise.all(asyncOperationsSnapshot);
+            } else {
+              // Promise rejections are allowed here, to be able to handle the error
+              // in <FeatureAppLoader />.
+
+              await Promise.allSettled(asyncOperationsSnapshot);
+            }
+          }
+
+          html = await render();
         }
-
-        html = await render();
-      }
-
-      while (asyncOperations.jsLoading.size > 0) {
-        while (asyncOperations.jsLoading.size > 0) {
-          inProgress = true;
-
-          const asyncOperationsSnapshot = Array.from(
-            asyncOperations.jsLoading.values(),
-          );
-
-          asyncOperations.jsLoading.clear();
-
-          // Promise rejections are allowed here, to be able to handle the error
-          // in <FeatureAppLoader />.
-
-          await Promise.allSettled(asyncOperationsSnapshot);
-        }
-
-        html = await render();
       }
     }
 
--- a/packages/async-ssr-manager/src/internal/async-ssr-manager.ts
+++ b/packages/async-ssr-manager/src/internal/async-ssr-manager.ts
@@ -114,6 +114,7 @@ export class AsyncSsrManager implements AsyncSsrManagerV1 {
         const asyncOperationsSnapshot = this.cloneQueue(asyncOperations);
 
         await Promise.all(asyncOperationsSnapshot.contentLoading);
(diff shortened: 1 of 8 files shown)


packages/async-ssr-manager/src/internal/async-ssr-manager.ts
```diff
@@ -50,29 +57,67 @@ export class AsyncSsrManager implements AsyncSsrManagerV1 {
...
+    // The `progress` flag acts as a sentinel: if no promise was processed
+    // during a full wave, the loop terminates (convergence).
 
-        const asyncOperationsSnapshot = Array.from(asyncOperations.values());
+    while (inProgress) {
+      inProgress = false;
 
-        asyncOperations.clear();
+      for (const queue of Object.keys(
+        asyncOperations,
```
stemey: Have you tried this solution:
```

while (isPromiseExisting(asyncOperations)) {
  while (isPromiseExisting(asyncOperations)) {

     asyncOperationsSnapshot = cloneQueue(asyncOperationsSnapshot)
     await Promise.all(asyncOperationsSnapshot.contentLoading);
     await Promise.allSettled(asyncOperationsSnapshot.jsLoading);
     
  }
  html = await render();

}

```

kmendzie: Hi @stemey yes I have tested the suggestion (https://github.com/feature-hub/feature-hub/pull/790/commits/296ae7f725b48cc4618207c309643df2bf20aeed). The `renderingLoop` looks good.

stemey (dismissed):

clebert (changes requested):
Kudos for tackling that complex topic :)

I have concerns about the approach taken here:

1. **API Design Issue**: The introduction of `_scheduleJsLoading` as a private API (indicated by the underscore prefix) is problematic. The method is named after a specific use case in the FeatureAppLoader, yet the AsyncSsrManager has no knowledge of loaders. This creates a coupling where the API design leaks implementation details of a consumer into a lower-level service.

2. **Simpler Alternative**: The core problem-ignoring errors when loading feature app JavaScript-can be solved locally in the FeatureAppLoader without modifying the AsyncSsrManager at all. Instead of:

```js
asyncSsrManager.scheduleRerender(loadingPromise);
```

Simply use:

```js
asyncSsrManager.scheduleRerender(loadingPromise.catch(() => {
  // Ignore error
}));
```

This handles the rejected promise at the source without requiring a new API.

The addition of `.idea` to `.gitignore` should be reconsidered. Personal IDE settings are typically better handled in a global gitignore rather than committed to the project configuration.

I'd suggest simplifying this PR by handling the error locally in the FeatureAppLoader and rolling back the AsyncSsrManager changes. This achieves the same goal with minimal code changes and no new API surface.

kmendzie:
Hey @clebert , 

Thank you for your feedback. Regarding point 2 (“Simpler Alternative”): I agree that this approach is appealing, but it would introduce a breaking change. Some existing implementations depend on the errors being thrown so they can handle these cases individually. By adding a `.catch(() => { /* ignore */ })`, those errors would no longer surface, removing that capability.

The alternative approach of introducing a semi-internal method, `_scheduleJsLoading`, would avoid a breaking change, as content-related promise rejections would still be thrown and remain catchable (cc @stemey).

From my perspective, we have two options going forward:

**A)** Accept the breaking change with ignoring errors and require consumers to catch errors individually (e.g. `<FeatureAppLoader onError={error => { throw error }} />`) to be aligned.

**B)** Along with point 2 (“Simpler Alternative”): Introduce a configuration option to avoid a breaking change. This aligns again with the intent originally described in the [bug report](https://github.com/feature-hub/feature-hub/issues/786#issue-2695040944) and slightly similar to this [old PR](https://github.com/feature-hub/feature-hub/pull/789).

I'd be glad to hear your feedback on this.

clebert:
Hey @kmendzie, thanks for the detailed reply and the follow-up commit.

Regarding the “breaking change” aspect: I think we’re actually in the same boat for both implementations.

* **Before** this PR, a failed `serverSrc` load scheduled via `asyncSsrManager.scheduleRerender(loadingPromise)` would cause `renderUntilCompleted` to reject, so integrators could catch the error at that level.
* In your *original* version with `_scheduleJsLoading`, the loading promise for the FeatureApp was moved into a `jsLoading` queue that is awaited via `Promise.allSettled`. From the perspective of `renderUntilCompleted`, those errors also no longer bubble out.
* In the *current* version (57af454), we instead pass `loadingPromise.catch(() => { /* ignore */ })` into `scheduleRerender`. That has the same effect: the promise seen by `AsyncSsrManager` never rejects, so `renderUntilCompleted` doesn’t throw for these module-load errors.

So for consumers using `<FeatureAppLoader>` together with `AsyncSsrManager.renderUntilCompleted`, both approaches change behaviour in the same way: module fetch failures no longer abort the whole SSR render, which is exactly what #786 asked for.

The difference is *only* where the special handling lives:

* With `_scheduleJsLoading`, AsyncSsrManager grows a semi-internal, loader-specific API and knows about this special case.
* With the current “simpler alternative”, the logic stays local to `FeatureAppLoader` and the AsyncSsrManager API remains untouched and generic.

For other consumers that call `scheduleRerender` directly, nothing changes in either variant: their promises are still awaited with `Promise.all` and rejections still cause `renderUntilCompleted` to reject.

Because of that, I don’t think `_scheduleJsLoading` actually avoids the “breaking” part you’re concerned about – it just introduces extra API surface without changing the observable behaviour compared to the local `.catch` solution.

Given that:

* I’d treat the old behaviour (a single failed module load killing the whole SSR) as a bug rather than an intentional contract.
* Integrators who **do** want to fail hard can still do so explicitly via `<FeatureAppLoader onError={error => { throw error; }} />`.
* We can document this change in the release notes as part of fixing #786.

I’m therefore 👍 on the current implementation (catching the error in `FeatureAppLoader` and passing a non-rejecting promise to `scheduleRerender`) and on keeping the AsyncSsrManager API as-is, without `_scheduleJsLoading` or additional configuration.

stemey (approved):

PR #793, opened 2026-06-22
build(all): migrate to npm

stemey:


stemey:
## Code Review: build(all): migrate to npm

### 🔴 High: pretest script installs Puppeteer browsers on every test run

**File:** [package.json](package.json#L25)

**Problem:** The `pretest` script runs `npm run install:puppeteer` which downloads Chromium browsers before **every** test execution. This adds unnecessary overhead to local development and should only run once during setup.

```json
"pretest": "rimraf coverage && npm run install:puppeteer"
```

**Fix:** Move `install:puppeteer` to be called only during initial setup (CI install step) or as an explicit prep command, not as a pre-hook for every test. Consider removing from pretest or creating a separate setup script that developers run manually:

```json
"setup": "npm install && npm run install:puppeteer",
"pretest": "rimraf coverage"
```

---

### 🟡 Medium: jest-puppeteer version jump (^9.0.0 → ^11.0.0) needs verification

**File:** [package.json](package.json#L66)

**Problem:** jest-puppeteer was bumped from ^9.0.0 to ^11.0.0, a major version jump. This introduces potential breaking changes that should be documented or tested.

**Fix:** Verify that all jest-puppeteer ^11.0.0 changes are compatible with the current test setup. Document the reason for this upgrade or test the full test suite before merge.

---

### 🟡 Medium: Documentation mentions lint script that is now disabled

**File:** [README.md](README.md#L43)

**Problem:** README still documents `npm run lint` command:
```md
- `npm run lint` — Lints all sources.
```

But [package.json](package.json#L24) now has:
```json
"lint": "echo \"No linting configured.\""
```

**Fix:** Either restore the lint implementation or update README to clarify that linting is not configured and remove the line entirely.

---

### 🟡 Medium: Node version matrix removed from CI

**File:** [.github/workflows/ci.yml](.github/workflows/ci.yml#L50-L53)

**Problem:** The test job previously ran matrix tests on Node 18 & 20. Now it runs only on the default version:

```yaml
steps:
  - name: Setup node ${{ matrix.node }}  # ${{ matrix.node }} is undefined
```

This reduces version compatibility testing coverage.

**Fix:** Either restore the node version matrix or document why single-version testing is acceptable:

```yaml
strategy:
  matrix:
    node: ['18', '20', '22']
```

---

### 🟢 Low: ci script uses --ignore-scripts flag

**File:** [package.json](package.json#L9)

**Problem:** The new `ci` script uses `--ignore-scripts`:
```json
"ci": "npm ci --prefer-offline --no-audit --no-fund --loglevel=error --ignore-scripts"
```

This prevents Husky install hook from running during `npm ci`. While intentional for CI environments, ensure this doesn't break local development workflow.

**Fix:** Document this design decision or provide a separate `setup` command for developers that includes scripts.

---

### ✅ Good Practices Observed

- ✓ Explicit permissions sections added to all workflows
- ✓ Actions upgraded to recent versions (checkout@v6, setup-node@v6)
- ✓ npm cache configuration with proper lockfile path
- ✓ Puppeteer browser caching added
- ✓ Documentation converted from yarn to npm consistently

stemey:
## Review Summary

- 🔴 High: 1
- 🟡 Medium: 3
- 🟢 Low: 1

**Key findings:** The npm migration is well-structured overall, but the pretest script change creates unnecessary Puppeteer browser downloads on every test run, and the jest-puppeteer major version jump needs verification before merge.

package.json
```diff
@@ -6,34 +6,35 @@
...
+    "ci": "npm ci --prefer-offline --no-audit --no-fund --loglevel=error --ignore-scripts",
     "clean": "tsc --build --clean && lerna exec --parallel -- rimraf lib",
-    "compile": "yarn compile:cjs && yarn compile:esm",
+    "compile": "npm run compile:cjs && npm run compile:esm",
     "compile:cjs": "tsc --build --pretty",
     "compile:esm": "lerna exec --no-private -- tsc --build tsconfig.esm.json",
-    "predeploy": "cdk bootstrap --app 'yarn aws-simple synth'",
-    "deploy": "cdk diff --app 'yarn aws-simple synth' && cdk deploy --app 'yarn aws-simple synth'",
+    "predeploy": "cdk bootstrap --app 'npm run aws-simple -- synth'",
+    "deploy": "cdk diff --app 'npm run aws-simple -- synth' && cdk deploy --app 'npm run aws-simple -- synth'",
```
chrbucs2, line 15: package.json:15 | 🔴 | `npm run aws-simple` cannot execute the `aws-simple` binary
Problem: `npm run` only invokes entries from the `scripts` section. Unlike Yarn, it does not forward unknown names to `node_modules/.bin`, so both `predeploy` and `deploy` will fail with `Missing script: "aws-simple"` before CDK even starts.

Code example:
```text
"predeploy": "cdk bootstrap --app 'npm run aws-simple -- synth'",
"deploy": "cdk diff --app 'npm run aws-simple -- synth' && cdk deploy --app 'npm run aws-simple -- synth'"
```

Fix: Call the binary directly inside the npm script (`aws-simple synth`) or use `npm exec`/`npx` instead of `npm run`.

.github/workflows/commands.yml
```diff
@@ -62,29 +67,31 @@ jobs:
...
+          cache-dependency-path: |
+            package-lock.json
 
       - name: Install
-        run: yarn install
+        run: npm run ci
 
       - name: Build Website
-        run: yarn lerna exec --scope @feature-hub/website 'docusaurus-build'
+        run: npm run lerna exec --scope @feature-hub/website 'docusaurus-build'
```
chrbucs2, line 85: .github/workflows/commands.yml:85 | 🔴 | `npm run lerna ...` breaks the deployment workflow
Problem: This uses the same Yarn-specific binary passthrough pattern as before, but `npm run` only executes named package scripts. Because there is no `lerna` script in the root `package.json`, this step exits with `Missing script: "lerna"`. The same issue also appears in the TodoMVC step here and in `deploy-main.yml`.

Code example:
```text
run: npm run lerna exec --scope @feature-hub/website 'docusaurus-build'
```

Fix: Execute the binary directly with `npx lerna ...` / `npm exec -- lerna ...`, or add dedicated npm scripts and call those from the workflow.

husky.config.js
```diff
@@ -2,6 +2,6 @@
 
 module.exports = {
   hooks: {
-    'commit-msg': 'yarn commitlint --edit',
+    'commit-msg': 'npm commitlint --edit',
```
chrbucs2, line 5: husky.config.js:5 | 🟡 | The commit-msg hook now points to an invalid npm command
Problem: `npm commitlint --edit` is not a valid npm invocation. `npm run` can execute scripts and `npm exec`/`npx` can execute binaries, but `npm <binary>` does neither here. That means local commit message validation stops working after the migration.

Code example:
```text
'commit-msg': 'npm commitlint --edit'
```

Fix: Switch to a binary execution form such as `npx commitlint --edit` or `npm exec -- commitlint --edit`.

package.json
```diff
@@ -6,34 +6,35 @@
...
     "postdeploy": "aws-simple upload --yes && aws-simple flush-cache --yes",
     "format": "prettier --write '**/*.{js,json,md,ts,tsx,yml}'",
     "generate:api-docs": "cross-env NODE_OPTIONS='--max-old-space-size=8192' typedoc --options typedoc.js .",
     "postgenerate:api-docs": "node scripts/process-api-docs.js",
-    "lint": "cross-env FORCE_COLOR=1 lerna exec --parallel -- tslint --config tslint.json --project tsconfig.json --format stylish",
+    "install:puppeteer": "puppeteer browsers install chrome",
+    "lint": "echo \"No linting configured.\"",
```
chrbucs2, line 21: package.json:21 | 🔴 | The lint quality gate was silently removed
Problem: This turns `npm run lint` into a no-op even though the repository still contains `tslint.json` files and the CI job is still named `Compile, Lint & Verify`. After this change, the workflow goes green without running any static analysis, so lint regressions are no longer caught.

Code example:
```text
"lint": "echo \"No linting configured.\""
```

Fix: Keep the existing TSLint command until a real replacement is ready, or migrate to ESLint in the same PR and update the CI/docs consistently.

.github/workflows/ci.yml
```diff
@@ -43,33 +48,37 @@ jobs:
...
+    name: Test
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Setup node ${{ matrix.node }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v6
```
chrbucs2, line 58: .github/workflows/ci.yml:58 | 🟡 | The workflow no longer pins a Node.js version
Problem: `setup-node` is now called without `node-version` or `node-version-file`, so the job falls back to whatever Node version happens to be preinstalled on the GitHub runner image. That makes CI less reproducible and can drift away from `devEngines.runtime.version` (`22.x`) without any change in this repository.

Code example:
```text
- name: Setup node ${{ matrix.node }}
  uses: actions/setup-node@v6
  with:
    cache: npm
```

Fix: Pin the intended runtime explicitly, e.g. `node-version: 22`, or keep a checked-in version file and reference it from all workflows.

chrbucs2 (commented):
Review summary

- ⚠️ Critical: 0
- 🔴 High: 3
- 🟡 Medium: 2
- 🟢 Low: 0

.github/workflows/ci.yml
stemey: node version is pinned by packageManager-engine.runtime

package.json
stemey: added "aws-simple" script

